PR #4, opened 2018-08-21
Avoid invoking ssh-agent and adding a key to it in favor of telling ssh where to find the private key.

jfly:
This feels a little simpler to me. Hopefully it works!

### DIFF
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -1,6 +1,4 @@
 #!/bin/bash
 
-eval "$(ssh-agent -s)" # Start ssh-agent cache
 chmod 600 /tmp/deploy_rsa # Allow read access to the private key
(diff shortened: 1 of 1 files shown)


jonatanklosko:
I considered that, but I followed [this blog post](https://oncletom.io/2016/travis-ssh-deploy/) and it suggested this way - as far as I understood a safer one, quoting:
> Adding the key to the current ssh-agent to make any SSH-based command agnostic to the private key location.

If it doesn't sound like a reason, I'm ok doing it easier ^^

jfly:
Ahh, I see. I guess I'd find that argument more compelling if we were doing multiple `ssh` or `rsync` invocations, or if we were providing a scrip that developers should run locally, but we're not doing either of those things.

That said, this is definitely not an important thing to change, and I think there's some value in being as similar as possible to an existing tutorial. I'll close this PR!

PR #11, opened 2018-08-27
Configure Travis to run on master branch only

jonatanklosko:


jonatanklosko:
Well, I consider Travis in this case as a deployment tool only, so I saw no reason for running it just to skip deployment steps. Most PRs are gonna change Markdown content, so no reason for `bin/build.sh` to fail; in case of PRs modifying the script itself, those are gonna be tested locally for sure. If you like running `bin/build.sh` anyway, I'm ok with that!

No need to worry about PR builds, they cannot decrypt the secret key file and thus fail (see [this build](https://travis-ci.org/thewca/wca-documents/builds/417583558) as an example). Given they would always fail, I disabled PR builds in the Travis config a while ago.

jfly:
I have a slight preference for running `build` but not `deploy` on PRs, but as you say, right now it probably doesn't matter at all. A couple of thoughts:

- I could totally see something breaking, and a lazy/quick developer just making some untested changes to `build.sh` to see if they can unbreak stuff.
- I could see us adding some "lint" steps to the build in the future. Maybe we'll want to make sure that there is no trailing whitespace, or no tabs, or maybe we'll do a spell or grammar check of some sort. In that case, we'd want to run some stuff in travis, but not the deploy step. That all said, we can totally change travis.yml if we ever want that behavior!

I've approved this PR, feel free to do whatever you think is best here!

jonatanklosko:
Makes sense, let's leave it as it is for now then!

PR #27, opened 2018-11-22
Motion Amendments 2019 - Senior Delegates

operge:
Fixes the issue of letting non-Full Delegates apply for Senior Delegate position.

### DIFF
--- a/documents/motions/09.2019.1 - Senior Delegates.md
+++ b/documents/motions/09.2019.1 - Senior Delegates.md
@@ -1,10 +1,10 @@
 # SUBMISSION OF PROPOSED MOTION
 
-**Motion number:** 9.2017.1  
+**Motion number:** 9.2019.1  
 **Subject:** Senior Delegates  
 **Intent:** Procedure of acknowledgement of WCA Senior Delegates and description of their rights and duties  
 **Submitted by:** Board of Directors  
-**Date:** July 1, 2017  
+**Date:** January 1, 2019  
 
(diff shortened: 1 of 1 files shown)


documents/motions/09.2017.1 - Senior Delegates.md
```diff
@@ -33,7 +33,7 @@ Senior Delegates are WCA Staff and Full Delegates, with an extra role to manage
       2. Perform as the escalation point for the WCA Delegates in case of issues with local Competition Organizers and Competition Staff.
 4. Appointment of Senior Delegates
    1. The Board of Directors decides how many Senior Delegates are needed and how the areas are divided among Senior Delegates, based on the number of delegates and the span-of-control of the Senior Delegates. All WCA Delegates should be managed by a Senior Delegate.
-   2. If a new Senior Delegate is needed for an area, the Board of Directors will request all WCA Staff Members and External Staff to apply as a candidate.
+   2. If a new Senior Delegate is needed for an area, the Board of Directors will request all Full Delegates to apply as a candidate.
```
AlbertoPdRF: This has always sounded weird to me, so while we are at it, why don't we reword it a bit?
```suggestion
   2. If a new Senior Delegate is needed for an area, the Board of Directors will ask for applicants among all Full Delegates.
```

AlbertoPdRF:
I've implemented the suggestion that was approved yesterday, so this one should be ready to be merged too!

Issue #5, opened 2018-08-21
A textual revision of the Motions is needed.


AlbertoPdRF:
Things to change:

- In [16.2017.1](https://github.com/thewca/wca-documents/blob/master/documents/motions/16.2017.1%20-%20Disputes.md) it says "Ethical Committee" when the official name is "Ethics Committee".
- In [8.2018.1](https://github.com/thewca/wca-documents/blob/master/documents/motions/08.2018.1%20-%20Delegates.md) it was introduced "with copy to the WCA Board" but should it be "with a carbon copy (CC) to the Board" as in [10.2017.0](https://github.com/thewca/wca-documents/blob/master/documents/motions/10.2017.0%20-%20Committees%20and%20Teams.md)?
- ...

cubewhiz:
Some more.

- In 4.2018.1 it says "processes that are lead by WCA Board" but it should be "led".
- In 1.2017.1, it should be a period and capital "S" for "See WCA Motion..."

PR #40, opened 2018-12-16
WCA Competition Requirements Policy v1.0

AlbertoPdRF:
This goes together with https://github.com/thewca/wca-regulations/pull/673


lgarron:
Checklist from https://github.com/thewca/wca-regulations/pull/673/files

Regs/Guidelines to move to the comp policy:

|Regulation/Guideline|To|
|----|----|
| 1c3+ and 1c4+ | 9.1.2 |
| 1c3++ | 9.1.2 |
|8a5+|8 (maybe)|

AlbertoPdRF:
By the way, I guess 1c3++) falls into the Data Protection scope, no @ChrisWrightWCA? I'm fine with adding it to the policy as is and amend it when we have a complete Data Protection policy though.

documents/policies/Competition Requirements.md
```diff
@@ -19,12 +19,15 @@ An official WCA Competition must:
 6. Be held on a single date or on two or more consecutive days.
 7. Have at least one official event.
 8. Have at least one competitor.
+   1. The competition should have at least 12 competitors.
```
cubewhiz: I think this is a worthless addition. What if it doesn't? We would still recognize the competition, anyway.

AlbertoPdRF: It's more like: "Aim to hold a competition where at least 12 competitors will participate" than anything else -- as it's a "should", I think this is fine. I'm happy to remove it though if that's the consensus!

lgarron: Indeed, I think "if you hold a competition then you should plan it so you get at least 12 competitors, else you better have a good reason" is still a good idea.

> 3. SHOULD   This word, or the adjective "RECOMMENDED", mean that there
>   may exist valid reasons in particular circumstances to ignore a
>   particular item, but the full implications must be understood and
>   carefully weighed before choosing a different course.
> (https://tools.ietf.org/html/rfc2119)

AlbertoPdRF:
Issues from the Regulations repository that haven't been handled yet:

- [ ] https://github.com/thewca/wca-regulations/issues/629
- [ ] https://github.com/thewca/wca-regulations/issues/641
- [ ] https://github.com/thewca/wca-regulations/issues/650

Jambrose777:
Adding solutions for #641 will also clear up that nobody should be registered for the competition prior to announcement.

Jambrose777:
I've made suggestions to the original document (google doc):

> + 6.2.2 The first registration must have a start time and date at least 48 hours after the competition is announced.

This clears up #641 and WCAT's concern I listed above.

Moved 5.3 to 6.2.3

AlbertoPdRF:
That proposal sounds good to me!

(By the way, I edited your comments to add the link to the issue.)

AlbertoPdRF:
That's a good point, and I agree with making such exceptions! Other opinions @thewca/board?

jfly:
> think having consistency in nicknames for these events is important for [...] database queries

For the record, people should *not* be using nicknames or ids to figure out the championship type of a competition. We include championship type information in the public database export, see https://docs.google.com/document/d/1tgIfNxJqYni_PFsMt-hpDSh1YpfoCseLORvLX6d7-l4/edit# and https://www.worldcubeassociation.org/results/misc/export.html for more information =)

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,90 @@
...
+
+If WCA Regulations are not correctly observed during a competition, results may be modified at the discretion of the WCA Delegate, WCA Regulations Committee, WCA Results Team, WCA Disciplinary Committee, or the WCA Board.
+
+### Announcement Criteria
+The WCA Delegate must submit the following information when requesting approval of a competition:
+
+1. Competition name, nickname, and ID
+   1. The competition name must abide by the [WCA Competition Name Policy](wca{documents/policies/Competition Name.pdf}).
+   2. The competition nickname and ID must resemble the competition name.
+      1. The competition nickname and ID should only be modified if the competition name is too long (more than 32 characters).
```
AlbertoPdRF:
```suggestion
       1. The competition nickname and ID should only be modified if the competition name is too long (more than 32 characters). Exception: World and Continental Championships should use nicknames and IDs that are consistent with past editions.
```

What about this @EdHollingdale?

Jambrose777:
Maybe make the exception more general. The idea here being a group of comeptitions where one isn’t he group exceeds the character limit
Take for instance comeptitions where the last word is the season.
Winter, spring, summer, fall.
The competition in the fall may not exceed the character limit while the others do, yet you may wish to have consistency with the nicknames and ids

AlbertoPdRF:
Maybe leaving it as it is is the best option after all, as a more general exception doesn't sound as good to me, and it's true that a "should" already leaves room for exceptions.

EdHollingdale:
Personally I'd prefer to have things written down in the requirements - especially if exceptions will only be granted in a limited number of circumstances (World and continental championships, competitions in a series or tour). This would make it easier for organisers and delegates to know when to request exceptions, and easier for WCAT to know when to grant the exceptions.

AlbertoPdRF:
Can you suggest the specific change so we can discuss about it better?

documents/policies/Competition Requirements.md
EdHollingdale:
```suggestion
      1. The competition nickname and ID should only be modified if the competition name is too long (more than 32 characters). Exception: Groups of competitions may have modified nicknames to ensure consistent nicknames with past and future competitions.
```

EdHollingdale: Not sure if change above is too general? I like having the exception for World and Continental championships explicitly called out, and I think having the consistency with WC_ _ _ _  is important. But I feel we should also allow exceptions for Jacob's example below (Fall/Winter/Summer/Spring competitions) and groups of competitions (e.g. West Coast Cubing Tour Fresno 2018 fits but West Coast Cubing Tour Los Angeles 2018 would not).

AlbertoPdRF: I'm ok with adding that sentence if others agree! But I would edit it to include IDs too.

EdHollingdale: I agree, the ID should also be covered with this.
```suggestion
      1. The competition nickname and ID should only be modified if the competition name is too long (more than 32 characters). Exception: groups of competitions may have modified nicknames and IDs to ensure consistency with past and future competitions.
```

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,90 @@
...
+      1. Qualification for an event must be based only on results from the same event.
+   2. If the competitor limit for a competition or an event is reached before the competition, the organization team must not approve any additional registrations for the competition or the event.
+   3. The competitor limit is per competition. Per day competitor limits are not allowed.
+6. Registration requirements
+   1. The registration requirements must be clear so that any competitor from anywhere in the world knows how to and has the ability to register.
+      1. The registration requirements must be available in English.
+      2. In addition to the English version, the registration requirements can be available in any other language and should be available in the local language.
+   2. The registration period(s) must be a clearly defined time range with a start and end time and date.
+      1. The first registration period must have a start time and date at least 48 hours after the competition is announced.
```
cubewhiz:
```suggestion
       1. The first registration period must have a start time and date at least 48 hours after the competition is announced.
          1. The competition's Delegate(s) and/or organizers may register for the competition prior to the first registration period.
```
Ideally, this would be automatic, but until then, I think allowing this manually (and then just accepting the registrations once the registration period opens) is best.

Jambrose777: For the record, wcat is strictly against this.

AlbertoPdRF: Against automatic registration of Delegates and/or organizers?

documents/policies/Competition Requirements.md
AlbertoPdRF: @thewca/board do you agree with the proposed wording by @EdHollingdale?

documents/policies/Competition Requirements.md
Jambrose777: Allowing them to register before everyone else

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,90 @@
...
+7. Events
+   1. Each event must have a planned number of rounds.
+      1. Each round of each event must have a planned format and time limit.
+      2. A round of an event must have planned advancement conditions if there is a subsequent round of the event.
+      3. A round may have a cutoff. A round with a cutoff is a combined round, with the formats as per [WCA Regulation 9b)](wca{regulations/#9b}).
+   2. Only results from events that appear on the competition website in the Events tab and Schedule will be recognized.
+      1. Other events may only be held unofficially.
+8. Schedule
+   1. The schedule must be published on the WCA website using the WCA schedule tool.
+9. Optional requirements
```
Jambrose777:
```suggestion
9. Championship type
    1. Championship types are for national, continental, and world championships.
    2. There should be no other competition of the same championship type in the same year.
    3. The championship has most of the officially recognized WCA events.
    4. The championship awards titles and awards to the winners within the respected region. 
    5. The organization team should involve the persons expected to be involved in the championship, otherwise those persons should be notified and satisfied with the organization team.
10. Optional requirements
```

I believe these are the 5 points of interest, I'm open ears to other points.
I feel we maybe should mention the naming policy with championships, but that would be redundant with 1.1.

5. is a little weirdly worded, if others have suggestions, I'm also all in :)

documents/policies/Competition Requirements.md
Jambrose777: But this is something that some competitions (or organizers or delegates) cannot control, where as typically everything else here is. I don't really see the point in adding the line.

cubewhiz: I also still prefer its removal. Perhaps write a line in the Crash Course suggesting that competitions should only be planned if you expect to have at least 12 competitors.

documents/policies/Competition Requirements.md
EdHollingdale: Maybe change 5. to "The organization team should consist of the Delegates of the respective region and the respective regional organization, otherwise those persons should be notified and satisfied with the members of the organization team" since I think that would cover the "expected persons". It doesn't really work for continental and world championships. Maybe add something about where a bidding process has taken place?

Jambrose777: Here is the new and improved wording:

```suggestion
9. Championship type
    1. A championship type is assigned to competitions that are national, continental, or world championships.
        1. A competition may have multiple championship types.
        2. An additional selection process is conducted by the WCA Board for organizing continental and world championships.
    2. A championship type for a region must only be assigned to one competition in a year.
    3. The competition should have most of the officially recognized WCA events.
    4. The competition should recognize champion titles and award them and their respective prizes to the winners in the region.
    5. The organization team of a national championship should involve all the Delegates of the respective country.
10. Optional requirements
```

AlbertoPdRF: I think the wording is good enough now. @thewca/board do you agree or is there anything else to change?

documents/policies/Competition Requirements.md
AlbertoPdRF: I'm fine with either keeping it or removing it, happy to go with the majority!

documents/policies/Competition Requirements.md
AlbertoPdRF: FTR: the Board consulted the WEC about this before making a final decision.

documents/policies/Competition Requirements.md
AlbertoPdRF: I edited a bit the improved wording. The only actual changes apart from styling are in points 3 ("most of" to "all", as the sentence uses a should) and 4 ("should" to "must"). I'll commit these changes soon if no one opposes!

AlbertoPdRF: Keep in mind that those points use "should", so reasons like the ones you describe are good enough to not follow the recommendation in such cases.

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,90 @@
...
+      1. Submitting the Delegate Report to the WCA Regulations Committee and the WCA Quality Assurance Committee regarding adherence to WCA Regulations and quality standards during the competition, the overall course of the competition, and any incidents that happened. The report must be submitted within one week of the end of the competition.
+      2. Sending the competition results to the WCA Results Team within one week of the end of the competition.
+         1. Results should be ready at the end of the last day of competition. Corrections to the results should be available within one week of the competition date.
+         2. All scramble sequences used during a competition must be sent with the results of the competition.
+            1. All scramble sequences must be labeled with the events, rounds, and groups for which they were used. If there are multiple groups on a round, it is not necessary to identify which competitor was in which group.
+         3. Physical records (e.g. score sheets, written solutions from 3x3x3 Fewest Moves) must be retained for a minimum of 90 days. Digital records (e.g. scramble sequences, results) must be retained permanently.
```
AlbertoPdRF: Should the part about keeping digital records permanently be amended? This will likely be required after a future Data Protection Policy is in force, so we might as well take care of it now!

documents/policies/Competition Requirements.md
jfly: @AlbertoPdRF, could you explain what that final decision was that the Board and the WEC came do? Could you also update https://github.com/thewca/worldcubeassociation.org/issues/1692 with the results of that decision? Thanks!

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,90 @@
...
+5. Be open to all competitors and spectators who wish to attend, up to the maximum capacity stated on the website or otherwise permitted by the venue.
+6. Be held on a single date or on two or more consecutive days.
+7. Have at least one official event.
+8. Have at least one competitor.
+   1. The competition should have at least 12 competitors.
+9. Have at least one designated WCA Delegate in attendance.
+   1. The WCA Delegate for a competition is responsible for:
+      1. Submitting the Delegate Report to the WCA Regulations Committee and the WCA Quality Assurance Committee regarding adherence to WCA Regulations and quality standards during the competition, the overall course of the competition, and any incidents that happened. The report must be submitted within one week of the end of the competition.
+      2. Sending the competition results to the WCA Results Team within one week of the end of the competition.
+         1. Results should be ready at the end of the last day of competition. Corrections to the results should be available within one week of the competition date.
```
SAuroux: This really doesn't reflect reality these days. Delegates are actually expected to double-check results before sending them in. Another soon to be merged commit will also introduce a major shift as several result checks will need to be passed before a result submission is even considered successful.

My Proposal: 
- "Submitting the competition results ..." instead of "Sending the competition results ..."
- Remove "Results should be ready at the end of the last day of competition. Corrections to the results should be available within one week of the competition date." 
- Add "Results should be double-checked and free of errors before submission." 
- Add "Corrections arising at a later point in time should be promptly sent to the WCA Results Team."

Jambrose777: image
I'm guessing this comment is not applicable anymore with a some point the highlighted text was added to the document?

cubewhiz: Indeed, I think the intentions of the document broadened. I agree with the suggestions made by @SAuroux.

AlbertoPdRF: I agree too!

documents/policies/Competition Requirements.md
operge: I am OK keeping it. Or how about saying that a competition should have a "reasonable amount of competitors"? This would cover cases where there is too few or when people do competitions for 30 when the demand is 200+. I know it's vague, but 12 is not a scientific number either.

documents/policies/Competition Requirements.md
operge: For 3 I agree, but what's the rationale behind 5? In some countries, the distances could be large, so why even suggest to drag them over there just for the sake of what? Also Bob's example is another thing. When we use "should" rules, we still try to aim to them, while having 50+ Delegates as an aim is not reasonable.

documents/policies/Competition Requirements.md
operge: Same here!

documents/policies/Competition Requirements.md
operge: I agree, it makes sense to change this now.

documents/policies/Competition Requirements.md
cubewhiz: If we wish to keep a line about a minimum number of competitors, I'd prefer something actionable so it can be evaluated. Example: "If a competition has fewer than 12 registered competitors at the end of the registration period or two weeks before the competition (whichever is earlier), the Delegate must contact the WCA Board." This way competitions with very few competitors can be evaluated to determine if they should still take place. Alternatively, I support removing the line completely.

documents/policies/Competition Requirements.md
cubewhiz: WCAT is against this and WEC is in favor. We are bringing the discussion to Leaders + Seniors now.

documents/policies/Competition Requirements.md
Jambrose777: I think the point of #5 is the opposite side. You shouldn’t have a foreign delegate conduct the national championship with a country of delegates.

The original read something like the persons expected to be involved in the organization teams hoild be.

documents/policies/Competition Requirements.md
AlbertoPdRF: The problem with "reasonable" is that different people will understand it differently, while with a fixed number we won't have that problem.

I'm fine with changing the sentence to something like what Bob mentioned.

documents/policies/Competition Requirements.md
AlbertoPdRF: @ChrisWrightWCA what do you think would be reasonable here?

documents/policies/Competition Requirements.md
AlbertoPdRF: Do you guys have any alternative wording in mind?

documents/policies/Competition Requirements.md
Jambrose777: I think my hold on adding what Bob said is that this is such a rare occurrence and will happen very rarely.
Here are the competitions with less than 12 competitors:
|Comp|competitors|
|--|--
|ChinaFMShenyang2011|3
|ChinaFMZhengzhou2011|3
|CubanOpen2016|6
|AkibaExpoOpen2018|6
|ChinaFMGuangzhou2011|6
|FantabulousMinnesota2015|6
|FantabulousMichigan2015|7
|BalakanQis2019|7
|IdealeNabeulOpen2018|7
|DutchOpen2003|7
|ValgaOpen2013|8
|LastMinuteQualifier2017|8
|TolongDiamMedan2018|8
|EdenVillageWinter2017|8
|Nakhchivan2018|8
|BeijingLongEvents2015|8
|RubikSinFronteras2015|9
|AvalonOpen2013|9
|SantiaguinodeVerano2014|9
|WaikatoAutumnOpen2010|9
|PBQUCLA2017|9
|PuyDeDomeOpen2010|9
|JordanOpen2012|9
|LagomarBast2018|9
|ChalmersOpen2005|9
|BogorOpen2013|9
|JogjaMini2015|10
|ArcticOpen2015|10
|MazateOpen2015|10
|Draughons2006|10
|AnotherBaliCubeDay2015|10
|MITFMC2014|10
|EstonianOpen2014|10
|BergshamraBlind2013|10
|SilencioTotal2018|10
|IdahoOpen2005|10
|GijonOpen2014|10
|CowboysFMCParty2017|10
|ChennaiMindOverMatter2018|11
|TartuOpen2016|11
|StadiFMC2018|11
|PleaseBeQuietWPI2017|11
|HeimaeyOpen2018|11
|OSUBlindWeekend2018|11
|JCubeJammu2017|11
|KotkaOpen2011|11
|MiniMinnesotaFall2017|11
|PleaseBeQuietBelgium2018|11
|Svekub2005|11
|IgnusOpen2016|11
|PleaseBeQuietReno2017|11
|SwisscubingFMC2018|11
|MalagaUIS2018|11

Would the board have denied any of these competitions from happening?

Ranzha:
To echo what Ron noted in the feedback email, it makes sense to incorporate the Competition Name and Competition Proximity policies directly in this document under "Announcement Criteria", under items 1 and 2 respectively.

documents/policies/Competition Requirements.md
AlbertoPdRF: After consultation with Leaders and Seniors, the Board has decided to allow Delegates and/or organizers to pre-register for their competitions, although their registrations must only be approved when registration opens.

documents/policies/Competition Requirements.md
AlbertoPdRF: Suggestion committed.

documents/policies/Competition Requirements.md
AlbertoPdRF:
```suggestion
         3. Physical records (e.g. score sheets, written solutions from 3x3x3 Fewest Moves) must be retained for a minimum of 90 days. Digital records (e.g. scramble sequences, results) must be retained for 7 years.
```
Following offline discussion.

AlbertoPdRF: The copy the Delegates' keeps yes, but of course not the actual data on our database :P Do you think this needs a clarification?

cubewhiz: yes

documents/policies/Competition Requirements.md
DanielEgdal: What about something like "at the discretion of the Senior Delegate"?

cubewhiz: I like that.

AlbertoPdRF: That sounds good to me too!

Ranzha: For point 4, perhaps:
`Championships must recognize and award champion titles and their respective prizes to the winners in the region.`
It's a weird sentence to de-clunk.

AlbertoPdRF: I'm fine with that suggestion. Others?

Jambrose777: Some new possible additions:

`The competition must be located in the region where the championship type is selected.`
Ie. You can have the European championship in America, or the Portuguese championship in Spain.

And then this one may require more discussion, and definitely needs better wording:
`A championship type cannot be on the same date as a conflicting championship type at a different competition.`
This would mean that Canada can't hold the Canadian Nationals on the same date as worlds or Euros and UK championship can't happen on the same weekend. This idea comes from the fact that in 2017 the Canadian nationals was held on the same weekend as worlds because "so someone other than Antoine and Bill can win". The last part of the sentence "different competitions" Is to ensure WC 2013 and 2019 can have the respective national championship at the same competition.

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,91 @@
...
+</div>
+
+## Purpose
+The purpose of this policy is to define the requirements and processes that a competition must meet and follow in order to be recognized as a WCA Competition.
+
+## Policy
+### WCA Competitions
+An official WCA Competition must:
+
+1. Be announced at least 28 days (four weeks) before the start of the competition on the WCA website.
```
cubewhiz:
```suggestion
1. Be announced at least 28 days (four weeks) before midnight of the start of the competition, measured in UTC, on the WCA website.
```
It should be clarified that we use the UTC time to determine the calendar and not the local time. Here's my attempt. This matches what the WCAT sees for __ days away when reviewing a competition.

documents/policies/Competition Requirements.md
cubewhiz: Probably not. All the more reason why we should get rid of this line and just add it to the crash course instead.

documents/policies/Competition Requirements.md
cubewhiz: For the first one, sure, that seems obvious.
The second one, I don't necessarily agree. If anything, maybe it can be a "should not." Otherwise, this also prevents WC 2021 from being announced on the same weekend of an already announced national championship. I think this point is too restrictive.

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,91 @@
...
+      7. Making a copy of the WCA Regulations available at the competition.
+   2. The WCA Delegate may delegate some of their responsibilities to organizers or other officials.
+
+If WCA Regulations are not correctly observed during a competition, results may be modified at the discretion of the WCA Delegate, WCA Regulations Committee, WCA Results Team, WCA Disciplinary Committee, or the WCA Board.
+
+### Announcement Criteria
+The WCA Delegate must submit the following information when requesting approval of a competition:
+
+1. Competition name, nickname, and ID
+   1. The competition name must abide by the [WCA Competition Name Policy](wca{documents/policies/Competition Name.pdf}).
```
cubewhiz: After hearing the feedback from Delegates, I agree that it is better to put the Competition Name policy here instead of in a separate document (though perhaps the "Championships" part of the Name policy belongs under "Championships" below.

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,91 @@
...
+   2. The competition nickname and ID must resemble the competition name.
+      1. The competition nickname and ID should only be modified if the competition name is too long (more than 32 characters). Exception: groups of competitions may have modified nicknames and IDs to ensure consistency with past and future competitions.
+2. Location(s)
```
cubewhiz: I think the proximity policy can fit under "Location(s)", though I think 3x3x3 Fewest Moves Simultaneous Competitions and Twin Competitions might not belong here. I am not sure the best place to describe those.

documents/policies/Competition Requirements.md
lgarron: Suggestion: "midnight preceding" is less ambiguous than "midnight of".

documents/policies/Competition Requirements.md
Jambrose777: But with a “should not” it would be reasonable to reject the national championship type on the same day of worlds with the worlds date being know, correct?

cubewhiz: In certain cases, yes. The further and poorer thr country is, though, the less reasonable it may be. For example, I think it's fine for Venezuela to have its national championship on the same day of Worlds in Australia, but New Zealand or United States would be less appropriate.

Jambrose777: So you’d also think the Canada situation described above is okay? Cause I certainly think it’s very cheeky.

cubewhiz: I think it should be avoided. Competitors from Canada are more likely to be able to afford the trip to Worlds. But with Euro, I think it's probably okay. The competitor would be making a choice to forfeit a national title opportunity to attend a competition in which they are not eligible for one. Should it be avoided? Yes. Should it be prohibited? No.

Jambrose777: Ah that's why I said conflicting. A competitor from Canada doesn't qualify as being able to be a european champion, so I wouldn't consider it conflicting. Whereas in the Canada/Worlds scenario the competitors have to choose which championship they'd rather be a part of. This feels very unfair to many competitors who now don't have a chance to even participate in the national championship every other year (assuming Canada will hold its nationals the same weekend as worlds from now on). This can be even worse for continents with championships. Let's say Hungary has a national championship that is always on the weekend of Euros or Worlds; then the top competitors from Hungary may never get the chance to participate in their own national championship since they are competing for the Euro/World champion titles.

documents/policies/Competition Requirements.md
Jambrose777: `3. Comply with all [WCA Competition Policies](wca{documents}).`
This line will need to be removed if so.

I feel like FMC competitions would also fit fine in the Locations section.

As well, we should probably define when a competition should be one competition instead of two (or vice versa) I.e. https://www.worldcubeassociation.org/competitions/PBQWaterloo2018 and https://www.worldcubeassociation.org/competitions/No3x3x3Waterloo2018 should have probably been 1 competition with 5km distance, disjoint events, same organizers/delegates, nearly same competitors. But where would we draw the line? WCAT briefly discussed this some time ago and never came to any conclusions about what is appropriate. Some felt that 5 km was too much here: https://www.worldcubeassociation.org/competitions/PeachState2019 So I'm not entirely convinced that there is a clear decision here, but would love to hear thoughts on it.

documents/policies/Competition Requirements.md
AlbertoPdRF: Ok, I'm removing the line from the new version I'm working on.

documents/policies/Competition Requirements.md
AlbertoPdRF: I'm adding both points on the new version I'm working on.

documents/policies/Competition Requirements.md
AlbertoPdRF: I'm adding this to the new version I'm working on, using 00:00 UTC to try not to be ambiguous.

documents/policies/Competition Requirements.md
AlbertoPdRF: I'm adding this to the new version too, as suggested.

documents/policies/Competition Requirements.md
AlbertoPdRF: I would keep 3. but removing the "Competition" from it. As for instance Dues Policy must still be observed, I think keeping the link is not a bad idea.

I have no good suggestion on when two competitions should be one :confused: 

By the way, I'm adding the whole Proximity Policy under "Location(s)", I will push the new version now.

cubewhiz: Perhaps when the organizer(s) & delegate(s) are the same and the competitions are on consecutive days in the same city.

AlbertoPdRF:
I force-pushed a new version including all feedback and squashing all commits into one.

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,141 @@
...
+   3. The name must end with the year of the last day of the competition.
+   4. The competition name must be in Roman-readable letters and may only contain characters from A-Z, a-z, and 0-9, with or without accents. Only '.:-& are permitted as special characters.
+   5. The competition name must be 50 characters or less, including whitespaces.
+   6. If the competition is named after anyone or anything, they must be persons or things lacking any controversial meaning worldwide.
+   7. Names should contain the location of the competition.
+   8. Competitions may not be named after a living cuber, competition organizer, or any competition staff.
+   9. Names must not contain possibly offensive parts or memes.
+   10. The name cannot contain website domains or anything susceptible to be interpreted as a link, hashtag or in general an electronic identifier.
+   11. The name cannot contain any trademarks (such as "Nike") unless the organization team is formally authorized, in writing, to use such names.
```
cubewhiz:
```suggestion
   11. The name cannot contain any trademarks unless the organization team is formally authorized, in writing, to use such names.
```

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,141 @@
...
+   12. Names should not contain any unnecessary parts in the name, like "competition".
+   13. The competition nickname and ID must resemble the competition name.
+       1. The competition nickname and ID should only be modified if the competition name is too long (more than 32 characters). Exception: groups of competitions may have modified nicknames and IDs to ensure consistency with past and future competitions.
+2. Location(s)
+   1. A location consists of the country, city, and venue.
+   2. The location should also include subregions of the country that are common to include with addresses (e.g. the state a U.S. city is located in).
+   3. Competitions will be accepted if there are no other competitions within 100 km driving distance and less than 19 days away.
+      1. Otherwise, the WCA Board should review the circumstances of the proposed competition. A Delegate requesting approval for such a competition must provide further information on why the competition should still be accepted. Possible arguments that support such a competition being accepted include:
```
cubewhiz:
```suggestion
      1. Otherwise, the WCA Competition Announcement Team should review the circumstances of the proposed competition. A Delegate requesting approval for such a competition must provide further information on why the competition should still be accepted. Possible arguments that support such a competition being accepted include:
```

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,141 @@
...
+1. Be announced at least 28 days (four weeks) before 00:00 UTC of the start date of the competition on the WCA website.
+   1. WCA Delegates should allow a suitable time frame for competitions to be announced after they are confirmed to the WCA Competition Announcement Team.
+2. Comply with the [WCA Regulations](wca{regulations}).
+3. Comply with all [WCA Policies](wca{documents}).
+4. Be publicly accessible.
+5. Be open to all competitors and spectators who wish to attend, up to the maximum capacity stated on the website or otherwise permitted by the venue.
+6. Be held on a single date or on two or more consecutive days.
+7. Have at least one official event.
+8. Have at least one competitor.
+9. Have at least one designated WCA Delegate in attendance.
```
Jambrose777: Apologies if this was already covered: should we remove "in attendance" as this is not true for 0.1% of competitions (estimated) and I would think that the board would wish to keep this exception for the 0.1% that have an emergency.

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,141 @@
...
+      1. The location of the competition
+      2. The competition venue
+      3. The season or date of the competition
+      4. The events of the competition
+      5. The convention the competition is a part of
+      6. A local cultural reference
+      7. A sponsor of the competition
+   2. Capitalization of the competition name must be appropriate. The first letter of each word in the competition’s name should be capitalized and subsequent letters within words should be lowercase, except articles and prepositions, which should be always lowercase. Acronyms should be capitalized.
+   3. The name must end with the year of the last day of the competition.
+   4. The competition name must be in Roman-readable letters and may only contain characters from A-Z, a-z, and 0-9, with or without accents. Only '.:-& are permitted as special characters.
```
Jambrose777:
```suggestion
   4. The competition name must be in Roman-readable letters and may only contain characters similar to A-Z, a-z, and 0-9. Only '.:-& are permitted as special characters.
```
otherwise even with the Roman-readable change, this would read that ł is not allowed.

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,141 @@
...
+   5. The competition name must be 50 characters or less, including whitespaces.
+   6. If the competition is named after anyone or anything, they must be persons or things lacking any controversial meaning worldwide.
+   7. Names should contain the location of the competition.
+   8. Competitions may not be named after a living cuber, competition organizer, or any competition staff.
+   9. Names must not contain possibly offensive parts or memes.
+   10. The name cannot contain website domains or anything susceptible to be interpreted as a link, hashtag or in general an electronic identifier.
+   11. The name cannot contain any trademarks (such as "Nike") unless the organization team is formally authorized, in writing, to use such names.
+   12. Names should not contain any unnecessary parts in the name, like "competition".
+   13. The competition nickname and ID must resemble the competition name.
+       1. The competition nickname and ID should only be modified if the competition name is too long (more than 32 characters). Exception: groups of competitions may have modified nicknames and IDs to ensure consistency with past and future competitions.
```
Jambrose777:
```suggestion
       1. The competition nickname and ID should only be modified if the competition name is too long (more than 32 characters). Exception: groups of competitions may have modified nicknames and IDs to ensure consistency with past and future competitions.
    14. Abbreviations and Acronyms should be avoided unless the competition name is too long (more than 32 characters).
```
This would also serve as "too generic", but I feel the extra clarification is good. Example: BU 2019 would not be accepted.

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,141 @@
...
+            1. both competitions should be at least 5 days apart, or
+            2. the other competition must not have the 3x3x3 Fewest Moves event.
+      2. A 3x3x3 Fewest Moves simultaneous competition cannot be part of any other regular competition.
+         1. Time frames when both are held cannot overlap.
+   5. Twin competitions
+      1. Twin competitions are WCA Competitions that are held less than 5 days apart for which the distance between them is less than 100 km, with the intention to accommodate more competitors to compete in one area.
+         1. Twin competitions may be accepted if the need for competitions in an area is so great that the community will benefit from having multiple competitions held in a short period of time.
+      2. Twin competitions are obliged to clearly and publicly state their condition of twin competitions and their special participation terms.
+         1. A competitor is only permitted to participate in one of the twin competitions.
+            1. If a competitor competes at more than one of the involved twin competitions, only the results from the competition that they compete in first will be kept.
```
Jambrose777: Has this ever been used? I think it's fine keeping it for someone to utilize in the future, just want to clarify that that hasn't happened yet.

I'd want to add the restriction that a competitor may not **register** for both competitions.

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,141 @@
...
+3. WCA Delegate(s)
+4. Organizer(s)
+   1. Organizers can be specific persons and/or organizations.
+5. Competitor limit
+   1. The competitor limit can be on either a "first come first serve" basis or based upon qualification results or rankings in the WCA world rankings of a previously announced calendar date.
+      1. Qualification for an event must be based only on results from the same event.
+   2. If the competitor limit for a competition or an event is reached before the competition, the organization team must not approve any additional registrations for the competition or the event.
+   3. The competitor limit is per competition. Per day competitor limits are not allowed.
```
Jambrose777:
```suggestion
   3. The competitor limit is per competition. Per day competitor limits are not allowed.
   4. The WCA Competition Announcement Team will review the competitor limit and/or qualification results or rankings to ensure that they are utilized properly.
```

Without something like this WCAT looses the ability to deny a competitor limit of 10 in NY, or a qualification time of 6 seconds in 3x3.

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,141 @@
...
+6. Registration requirements
+   1. The registration requirements must be clear so that any competitor from anywhere in the world knows how to and has the ability to register.
+      1. The registration requirements must be available in English.
+      2. In addition to the English version, the registration requirements can be available in any other language and should be available in the local language.
+   2. The registration period(s) must be a clearly defined time range with a start and end time and date.
+      1. The first registration period must have a start time and date at least 48 hours after the competition is announced.
+         1. The competition's Delegate(s) and/or organizers may register for the competition prior to the first registration period.
```
Jambrose777:
```suggestion
         1. The competition's WCA Delegate(s) and/or organizer(s) may register for the competition prior to the first registration period, however must not be approved until the first registration period has opened.
```

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,141 @@
...
+      2. If there is more than one registration period, the organization team must announce each additional registration period at least 48 hours before it starts.
```
Jambrose777: 1. and 2. can be combined:
```suggestion
      1. Each registration period must be announced publicly on WCA at least 48 hours before it starts.
```
But I'm also fine keeping it the way it is now for clarity.

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,141 @@
...
+      1. Continental and World Championships names must begin with "WCA".
+      2. The word "Championship" must be singular in the competition name.
+      3. Competitions which are not National, Continental, or World championships should avoid names that are similar to championship names.
+   4. Date(s)
+      1. National Championships should not be held on the same date(s) as Continental or World championships.
+      2. Continental Championships must not be held on the same date(s) as World Championships.
+   5. Championships should have all the officially recognized WCA events.
+   6. Championships must recognize and award champion titles and their respective prizes to the winners in the region.
+   7. Organization teams of National Championships should involve all the Delegates of the respective countries, at the discretion of the Senior Delegate.
+10. Optional requirements
```
Jambrose777:
```suggestion
10. Optional requirements (Approval from the WCA Competition Announcement Team must be given)
```

or something of the sort.

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,141 @@
...
+   3. Different locations for multi-location competitions may have different registration requirements (e.g. competitor limit).
+7. Events
+   1. Each event must have a planned number of rounds.
+      1. Each round of each event must have a planned format and time limit.
+      2. A round of an event must have planned advancement conditions if there is a subsequent round of the event.
+      3. A round may have a cutoff. A round with a cutoff is a combined round, with the formats as per [WCA Regulation 9b)](wca{regulations/#9b}).
+   2. Only results from events that appear on the competition website in the Events tab and Schedule will be recognized.
+      1. Other events may only be held unofficially.
+8. Schedule
+   1. The schedule must be published on the WCA website using the WCA schedule tool.
```
Jambrose777:
```suggestion
   1. The schedule must be published on the WCA website using the WCA schedule tool.
   2. The schedule should be manageable by the competition's WCA Delegate(s) and adhere to the WCA quality standards.
```

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,141 @@
...
+9. Championships
+   1. A championship type is assigned to competitions that are National, Continental, or World Championships.
+      1. A competition must be located in the region where the championship type is selected.
+      2. A competition may have multiple championship types.
+      3. A championship type for a region must only be assigned to one competition in a year.
+      4. An additional selection process is conducted by the WCA Board for organizing Continental and World Championships.
+   3. Naming
```
Jambrose777:
```suggestion
   2. Naming
```
And renumbering down.

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,141 @@
...
+      1. Continental and World Championships names must begin with "WCA".
+      2. The word "Championship" must be singular in the competition name.
```
Jambrose777:
```suggestion
      2. If used, the word "Championship" must be singular in the competition name.
```

Otherwise, German Nationals and German Open would not be acceptable names for a championship.

documents/policies/Competition Requirements.md
Jambrose777:
```suggestion
      1. Otherwise, the WCA Board should review the circumstances of the proposed competition. The WCA Delegate requesting approval for such a competition must provide further information on why the competition should still be accepted. Possible arguments that support such a competition being accepted include:
```

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,141 @@
...
+      3. Competitions which are not National, Continental, or World championships should avoid names that are similar to championship names.
+   4. Date(s)
+      1. National Championships should not be held on the same date(s) as Continental or World championships.
+      2. Continental Championships must not be held on the same date(s) as World Championships.
+   5. Championships should have all the officially recognized WCA events.
+   6. Championships must recognize and award champion titles and their respective prizes to the winners in the region.
+   7. Organization teams of National Championships should involve all the Delegates of the respective countries, at the discretion of the Senior Delegate.
```
Jambrose777:
```suggestion
   7. Organization teams of National Championships should involve all the WCA Delegates of the respective countries, at the discretion of the Senior Delegate.
```

documents/policies/Competition Requirements.md
cubewhiz: Perhaps it makes sense to add this (stolen from Motion 8.2019.1):
Exception: in case the WCA Delegate cannot attend the competition, the WCA Board may
appoint a trusted person to act as a temporary WCA Candidate Delegate for the
competition.

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,141 @@
...
+## Policy
+### WCA Competitions
+An official WCA Competition must:
+
+1. Be announced at least 28 days (four weeks) before 00:00 UTC of the start date of the competition on the WCA website.
+   1. WCA Delegates should allow a suitable time frame for competitions to be announced after they are confirmed to the WCA Competition Announcement Team.
+2. Comply with the [WCA Regulations](wca{regulations}).
+3. Comply with all [WCA Policies](wca{documents}).
+4. Be publicly accessible.
+5. Be open to all competitors and spectators who wish to attend, up to the maximum capacity stated on the website or otherwise permitted by the venue.
```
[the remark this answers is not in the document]
cubewhiz: I think 4 is a subset of 5. (If 4 was omitted, would we lose anything)?

documents/policies/Competition Requirements.md
cubewhiz: I would prefer to keep the A-z, a-z, 0-9 and add "or similar characters" after, then. The suggestion could be interpreted that the original letters A-Z are not permitted--only letters similar to them. (Yes, I know it's a ridiculous interpretation.)

documents/policies/Competition Requirements.md
cubewhiz: Agreed.

documents/policies/Competition Requirements.md
cubewhiz: It was *almost* used last year (Sam and I almost both delegated competitions in NYC on the same date). I think there may be cases that exist before I joined the Board.

I think it's reasonable that they can only register for one (but I think it's also reasonable that they can change their registration if a second competition is announced when they're already registered for the first). In an ideal world, our software handles this all cleanly, but it's a lot of work for something that is almost never used.

documents/policies/Competition Requirements.md
cubewhiz: Do you think Motion 10.2019.1 is too broad?
2. The WCA Competition Announcement Team has the following rights and duties:
2.1 Announcing WCA Competitions:
2.1.1 Approving and announcing WCA Competitions.
2.1.1.1 Ensuring announcements adhere to WCA quality standards.
2.1.2 Rejecting competitions.

documents/policies/Competition Requirements.md
Jambrose777:
```suggestion
   4. The competition name must be in Roman-readable letters and may only contain characters from A-Z, a-z, 0-9, or similar characters. Only '.:-& are permitted as special characters.
```

documents/policies/Competition Requirements.md
cubewhiz: If a change is made about this, I'd prefer it be in 10.2019.1 instead of this policy, but I think 2.1.1.1 reasonably allows rejection of the examples you described.

documents/policies/Competition Requirements.md
cubewhiz: I think this is already explicit from 10.2019.1:
2.1.4 Approving the adoption of optional Regulations by organization teams to facilitate
the administration of competitions.

cubewhiz: If you think a change is due, perhaps it can be 2.1.4 Approving (or rejecting)... in the Motion.

documents/policies/Competition Requirements.md
Jambrose777: Hmmm, reading the document as a whole without the comments, I think that the document makes it sound like these items (and the optional requirements) will be accepted without any pushback from WCAT, whenever we would not accept things, most of the time related in these areas, due to them not being acceptable.

cubewhiz:
@thewca/board please review to see if you think there need to be any other changes.

documents/policies/Competition Requirements.md
EdHollingdale: Perhaps add a condition that the twin competitions be announced jointly, or that registration for the first does not open until the second is announced? I feel that would make it fairer and remove the difficulty of making it very simple to switch registrations. 

I could see this become relevant in areas where the effort of organising local competitions for 200+ competitors becomes too much - it could be cheaper and less work to organise two sets of 100 person competitions.

documents/policies/Competition Requirements.md
EdHollingdale: I agree with Jacob here. From a delegate's and organisers perspective, it is better to have all of the requirements in one place, rather than having to cross reference the motions and the policy.

documents/policies/Competition Requirements.md
EdHollingdale: Seeing as this seems to be standard practice for WCAT it would be beneficial to have it written down somewhere.

Perhaps we should add another exception to the competition ID rule so that it may contain the acronym version to improve searchability on the WCA website - e.g. Please Be Quiet Oshawa 2019 can easily be found if I type in "PBQ Oshawa" into the search box, but searching for "PBQ Korea" gives nothing.

Jambrose777: I’d disagree with that. If we want to increase searcability, we should add functionality to the search feature itself. 
A competition should only have 1 name. Right now there are 3 boxes used for the competition name, allowing for 3 different names to be used.
If we want to increase search ability, why stop at 3 names, why not give the competition 100 different searchable names?

documents/policies/Competition Requirements.md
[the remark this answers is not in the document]
henrikthedane: I do not think we would lose anything.

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,141 @@
...
+   1. The WCA Delegate for a competition is responsible for:
+      1. Submitting the Delegate Report to the WCA Regulations Committee and the WCA Quality Assurance Committee regarding adherence to WCA Regulations and quality standards during the competition, the overall course of the competition, and any incidents that happened. The report must be submitted within one week of the end of the competition.
+      2. Submitting the competition results to the WCA Results Team within one week of the end of the competition.
+         1. Results should be double-checked and free of errors before submission.
+         2. All scramble sequences used during a competition must be sent with the results of the competition.
+            1. All scramble sequences must be labeled with the events, rounds, and groups for which they were used. If there are multiple groups on a round, it is not necessary to identify which competitor was in which group.
+         3. Physical records (e.g. score sheets, written solutions from 3x3x3 Fewest Moves) must be retained for a minimum of 90 days. Digital records (e.g. scramble sequences, results) must be retained for 7 years.
```
pokekrom: Aren't scrambles stored in the WCA database?
Do delegates have to keep the scrambles/results file in their computer for 7 years?

Jambrose777: Yes. What if the wca database is deleted?
Btw it doesn’t have to be on your local computer. This can also be possible via maintaining the results email thread (ie just don’t delete it)

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,141 @@
...
+      1. Otherwise, the WCA Board should review the circumstances of the proposed competition. A Delegate requesting approval for such a competition must provide further information on why the competition should still be accepted. Possible arguments that support such a competition being accepted include:
+         1. The two competitions don’t have events in common.
+         2. The two competitions take place in different countries.
+   4. 3x3x3 Fewest Moves simultaneous competitions
+      1. 3x3x3 Fewest Moves simultaneous competitions are WCA competitions having only the 3x3x3 Fewest Moves event and taking place simultaneously in different locations.
+         1. If any location of such competitions is closer than 100 km to any other competition, then
+            1. both competitions should be at least 5 days apart, or
+            2. the other competition must not have the 3x3x3 Fewest Moves event.
+      2. A 3x3x3 Fewest Moves simultaneous competition cannot be part of any other regular competition.
+         1. Time frames when both are held cannot overlap.
```
DanielEgdal: I believe around here we should add something about the deadline for adding a venue/location to the Fewest Moves Simultaenous competitions. WCAT already mentioned in the Crash-course that "The latest deadline to add a new location is 28 days before the competition."

cubewhiz: Sure, that's reasonable.

documents/policies/Competition Requirements.md
EdHollingdale: I agree that a competition should have one name. But teaching a search function a thousand different acronyms is hard. Having an ID reflect a shorter version of that one competition name is easier. I agree we should avoid acronyms in official names where possible, but we should recognise that acronyms and shortenings do exist for convenience (even when the full text is less that 32 characters).

For example, Melbourne had St Kevin's College Spring 2018, but everyone called it "SKC Spring". If someone asks "are you going to register for SKC?" We should make it as easy as possible for that competitor to find the competition their friend is talking about.

Jambrose777: Hello Ed, I'll summarize a conversation with the UK delegates here:
- You are asking for a search capability, not something the ID of a competition should stand for. The ID is a unique field used in the urls and data identifiers for the competition. It is just the name of the competition without spaces. If you want to increase search capabilities that's something different that should be implemented elsewhere. It certainly doesn't require writing thousands of patterns of code to produce. Another idea would be to add searchable tags for a competition (similar to Youtube tags) where if you search for those keywords, it will pop up in the search box. Again this is a different feature unrelated to the id.
- By having the ID be different, we are essentially creating multiple names for a singular competition. Why stop at just 3 (name, nickname, and ID), why can't I have 100 names for a competition? the idea here is to make sure that there are as few officially recognized names as possible for a competition (ideally 1, but 2 if it is over 32 characters).
- Shortening the ID to an acronym heightens the possibility of duplicates among them. (IE China Championship 2019 and Classic City 2019 would want the same CC2019 id). Why would it be fair to let one competition have this over another. (of note: yet, this can be an issue for strait up claiming a name. But from what I've seen delegates have been rather respectful and not taking away names from one another).
- We are wanting more worldwide consistency. So If we have one competition have the name be an acronym, we should probably have them all be like this. You run into the issue above and this just becomes a mess.
- Readability becomes an issue as SKC2019, we have no idea what this competition is for or even what competition name that it relates to. Overall, this just makes a list of Ids a list of nonsense: Ags2014, SCMU2015, SIGI2015, GBCO2015, PoP2015, GIC2016, OFFA2016, SCMU2016, ABHC2016 theres about 50 names like these in the database right now which you cant really make sense of.

EdHollingdale: Thanks for the explanation Jacob. I agree that the best solution is probably tags or similar in the website. I'm certainly not supportive of multiple names for a competition - there should be one official name. But there should also be scope to allow people to easily find the competition when they know it as only a simplification of the official name. Currently the easiest way to do that is through the competition ID. Given ambiguity, the potential for duplicates etc. we should use tags or similar instead as a long term solution. I feel in the mean time WCAT should perhaps be more generous with allowing different IDs in some cases when they contain well known acronyms that can ease search ability (e.g. ABHC, PBQ).

cubewhiz:
Are there any concerns still pending?

Jambrose777:
> Are there any concerns still pending?

I think the items about WCAT rejecting stuff is covered by the motions as you said. Although for this one: https://github.com/thewca/wca-documents/pull/40#discussion_r260113795 I think it is best to make these items explicitly require approval from the committee.

Other than that the document just needs to be changed on the items that the board agrees with :)

documents/policies/Competition Requirements.md
cubewhiz: How about we do both (change the motion and make it explicit here)?

Jambrose777: I'd be good with that!

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,141 @@
...
+         1. Twin competitions may be accepted if the need for competitions in an area is so great that the community will benefit from having multiple competitions held in a short period of time.
+      2. Twin competitions are obliged to clearly and publicly state their condition of twin competitions and their special participation terms.
+         1. A competitor is only permitted to participate in one of the twin competitions.
+            1. If a competitor competes at more than one of the involved twin competitions, only the results from the competition that they compete in first will be kept.
+3. WCA Delegate(s)
+4. Organizer(s)
+   1. Organizers can be specific persons and/or organizations.
+5. Competitor limit
+   1. The competitor limit can be on either a "first come first serve" basis or based upon qualification results or rankings in the WCA world rankings of a previously announced calendar date.
+      1. Qualification for an event must be based only on results from the same event.
```
EdHollingdale: As @Jambrose777 suggested [in my suggestion about Z4+](https://github.com/thewca/wca-regulations/issues/719) it would be good to have that clarity in this document, since the optional regulations will live here. 

Maybe we could have something like:
```suggestion
      1. Qualification for an event must be based only on results from the same event.
      2. Qualification for an event should be based on one type of result (e.g. average/mean or single) with the second type of result used only to distinguish between competitors with an identical position using the preferred result.
```
Wording of the second part is quite clunky however - happy for someone to suggest something better! The core idea is that qualification results should be clear from the start (so people know what they need to achieve) and without ambiguity (e.g. it should be clear if it is average or single).

documents/policies/Competition Requirements.md
AlbertoPdRF: I'm merging both suggestions into Bob's and commiting it.

documents/policies/Competition Requirements.md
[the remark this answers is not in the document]
AlbertoPdRF: I will get rid of 4 and keep 5 as the new 4.

documents/policies/Competition Requirements.md
AlbertoPdRF: Sounds good to me!

```suggestion
9. Have at least one designated WCA Delegate in attendance.
   1. Exception: in case the WCA Delegate cannot attend the competition, the WCA Board may
appoint a trusted person to act as a temporary WCA Candidate Delegate for the
competition.
```

documents/policies/Competition Requirements.md
AlbertoPdRF: Has an issue about this feature requests been open in https://github.com/thewca/worldcubeassociation.org?

Related to the suggestion: I'm editing it to change the `A` of `Acronyms` to a lowercase `a` and commiting it.

documents/policies/Competition Requirements.md
AlbertoPdRF: Specific suggested changes are appreciated if someone feels strongly about changing this.

documents/policies/Competition Requirements.md
AlbertoPdRF: That sounds reasonable to me too.

documents/policies/Competition Requirements.md
AlbertoPdRF: I'm adding a `their registration(s)` and commiting this suggestion too!

documents/policies/Competition Requirements.md
AlbertoPdRF: I think that's a good suggestion, I will add it when pushing all the changes!

documents/policies/Competition Requirements.md
AlbertoPdRF: I think this would look better as a sub-point instead of as it is suggested now, no?

documents/policies/Competition Requirements.md
AlbertoPdRF: Ups, good catch! I'll take care of this too.

documents/policies/Competition Requirements.md
AlbertoPdRF: If you feel like suggesting a change directly, that would be appreciated! :D

documents/policies/Competition Requirements.md
AlbertoPdRF:
```suggestion
      1. Qualification for an event must be based only on results from the same event.
      2. Qualification for an event must be based on one type of result only (i.e. average/mean or single)
         1. Exception: in case of using average/mean results, single results must be used to undo ties.
```

Does this look clearer?

AlbertoPdRF:
I force-pushed a new version which includes all suggested changes except for:

- https://github.com/thewca/wca-documents/pull/40#discussion_r260111259
- https://github.com/thewca/wca-documents/pull/40#discussion_r260113795
- https://github.com/thewca/wca-documents/pull/40#discussion_r261367158
- https://github.com/thewca/wca-documents/pull/40#discussion_r271763465

documents/policies/Competition Requirements.md
EdHollingdale: This looks good @AlbertoPdRF - I think having it as a sub-point is definitely the way to go. Do we want it to say "undo ties" or "break ties"?

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,145 @@
...
+         1. The competition's WCA Delegate(s) and/or organizer(s) may register for the competition prior to the first registration period, however their registration(s) must not be approved until the first registration period has started.
+      2. Different registration periods may have different registration requirements (e.g. restricting which events a competitor can register for).
+   3. Different locations for multi-location competitions may have different registration requirements (e.g. competitor limit).
+7. Events
+   1. Each event must have a planned number of rounds.
+      1. Each round of each event must have a planned format and time limit.
+      2. A round of an event must have planned advancement conditions if there is a subsequent round of the event.
+      3. A round may have a cutoff. A round with a cutoff is a combined round, with the formats as per [WCA Regulation 9b)](wca{regulations/#9b}).
+   2. Only results from events that appear on the competition website in the Events tab and Schedule will be recognized.
+      1. Other events may only be held unofficially.
```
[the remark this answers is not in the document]
AlbertoPdRF: I agree with that addition. Two comments, though:

1. Should we change the `should` to a `must`?
2. The indentation looks wrong.

documents/policies/Competition Requirements.md
AlbertoPdRF: Probably "break ties" sounds better? I'll let a native speaker reply to that question :stuck_out_tongue:

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,145 @@
...
+2. Location(s)
+   1. A location consists of the country, city, and venue.
+   2. The location should also include subregions of the country that are common to include with addresses (e.g. the state a U.S. city is located in).
+   3. Competitions will be accepted if there are no other competitions within 100 km driving distance and less than 19 days away.
+      1. Otherwise, the WCA Competition Announcement Team should review the circumstances of the proposed competition. The WCA Delegate requesting approval for such a competition must provide further information on why the competition should still be accepted. Possible arguments that support such a competition being accepted include:
+         1. The two competitions don’t have events in common.
+         2. The two competitions take place in different countries.
+   4. 3x3x3 Fewest Moves simultaneous competitions
+      1. 3x3x3 Fewest Moves simultaneous competitions are WCA competitions having only the 3x3x3 Fewest Moves event and taking place simultaneously in different locations.
+         1. If any location of such competitions is closer than 100 km to any other competition, then
```
DanielEgdal:
```suggestion
         1. Like for a regular competition, the deadline for adding a new location to a 3x3x3 Fewest Moves simultaneous competition is 28 days. 
         2. If any location of such competitions is closer than 100 km to any other competition, then
```

documents/policies/Competition Requirements.md
DanielEgdal: I did slightly above. Feel free to change wording as seen fit.

AlbertoPdRF: Thanks! I'm resolving this conversation in favor of the new one :smile:

documents/policies/Competition Requirements.md
AlbertoPdRF: Maybe `Any location must be announced at least 28 days (four weeks) before 00:00 UTC of the start date of the competition on the WCA website.` is better for consistency?

documents/policies/Competition Requirements.md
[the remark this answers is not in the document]
Jambrose777: > I agree with that addition. Two comments, though:
> 
> 1. Should we change the `should` to a `must`?
> 2. The indentation looks wrong.

I think using 'should' will be good right now. This gives a bit of grace period for delegates who have the potential to make a mistake. I wouldn't consider it right to deny a main event (which realistically only hurts the podium of the main event) due to this mistake at this point in time. I think it may be something to change at a later version (maybe 6months to a year from this document's announcing).

documents/policies/Competition Requirements.md
Jambrose777: Break ties is correct.

documents/policies/Competition Requirements.md
cubewhiz: Sure, that sounds reasonable.

lgarron:
The WCA Regulations are going into effect May 1: https://www.worldcubeassociation.org/posts/wca-regulations-may-2019

While I don't think there is a problem in practice if the competition requirements policy is not ready, it would help make  sure that every detail is continuously covered by a full-fledged policy.

cubewhiz:
> The WCA Regulations are going into effect May 1: https://www.worldcubeassociation.org/posts/wca-regulations-may-2019
> 
> While I don't think there is a problem in practice if the competition requirements policy is not ready, it would help make sure that every detail is continuously covered by a full-fledged official document.

That was the primary reason I suggested May 1 as the implementation date for the new Regs. We should be able to finalize this by then.

AlbertoPdRF:
We are almost there actually -- I'll push yet another revision of the policy now and it should be ready for a final review.

documents/policies/Competition Requirements.md
AlbertoPdRF: I'm resolving this conversation now, let's discuss this for version 1.1 :stuck_out_tongue:

documents/policies/Competition Requirements.md
AlbertoPdRF: Changed to `break ties`, I'm committing this!

documents/policies/Competition Requirements.md
AlbertoPdRF: I edited the original suggestion according to my comment and I'm committing it!

documents/policies/Competition Requirements.md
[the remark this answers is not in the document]
AlbertoPdRF: Thanks, LGTM now! I'm committing it as is.

documents/policies/Competition Requirements.md
AlbertoPdRF: I'm pushing a new version now with this change included!

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,150 @@
...
+5. Competitor limit
+   1. The competitor limit can be on either a "first come first serve" basis or based upon qualification results or rankings in the WCA world rankings of a previously announced calendar date.
+      1. Qualification for an event must be based only on results from the same event.
+      2. Qualification for an event must be based on one type of result only (i.e. average/mean or single)
+         1. Exception: in case of using average/mean results, single results must be used to break ties.
+   2. If the competitor limit for a competition or an event is reached before the competition, the organization team must not approve any additional registrations for the competition or the event.
+   3. The competitor limit is per competition. Per day competitor limits are not allowed.
+   4. The WCA Competition Announcement Team will review the competitor limit and/or qualification results or rankings to ensure that they are utilized properly.
+6. Registration requirements
+   1. The registration requirements must be clear so that any competitor from anywhere in the world knows how to and has the ability to register.
```
SAuroux: Ouch, this sentence definitely has to change, as it is literally impossible to fulfill.

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,150 @@
...
+      3. Competitions which are not National, Continental, or World championships should avoid names that are similar to championship names.
+   3. Date(s)
+      1. National Championships should not be held on the same date(s) as Continental or World championships.
+      2. Continental Championships must not be held on the same date(s) as World Championships.
+   4. Championships should have all the officially recognized WCA events.
+   5. Championships must recognize and award champion titles and their respective prizes to the winners in the region.
+   6. Organization teams of National Championships should involve all the WCA Delegates of the respective countries, at the discretion of the Senior Delegate.
+10. Optional requirements
+    1. Require competitors to submit puzzles at a specific time before an event is scheduled to begin.
+    2. Prohibit competitors from participating in specific combinations of events.
```
SAuroux: 10.1 and 10.2 should be full sentences (like every other paragraph before).

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,150 @@
...
+   1. Each event must have a planned number of rounds.
+      1. Each round of each event must have a planned format and time limit.
+      2. A round of an event must have planned advancement conditions if there is a subsequent round of the event.
+      3. A round may have a cutoff. A round with a cutoff is a combined round, with the formats as per [WCA Regulation 9b)](wca{regulations/#9b}).
+   2. Only results from events that appear on the competition website in the Events tab and Schedule will be recognized.
+      1. Other events may only be held unofficially.
+   3. If the main event of the competition is not the 3x3x3 Cube, it should be publicly mentioned in the competition's information.
+8. Schedule
+   1. The schedule must be published on the WCA website using the WCA schedule tool.
+   2. The schedule should be manageable by the competition's WCA Delegate(s) and adhere to the WCA quality standards.
```
SAuroux: "WCA quality standards" is mentioned here and in the "first" 8.2.1, but what are those? Such an expression requires a definition in my view.

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,150 @@
+# WCA Competition Requirements Policy
+<div class="version">
+### Version 1.0
+</div>
+
+## Purpose
+The purpose of this policy is to define the requirements and processes that a competition must meet and follow in order to be recognized as a WCA Competition.
+
+## Policy
+### WCA Competitions
```
SAuroux: The second paragraph is named "Announcement Criteria" and contains criteria. This paragraph is named "WCA Competitions" but obviously doesn't contain competitions ;). Long story short: This should be named "WCA Competition criteria" or "WCA Competition requirements".

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,150 @@
...
+   10. The name cannot contain website domains or anything susceptible to be interpreted as a link, hashtag or in general an electronic identifier.
+   11. The name cannot contain any trademarks unless the organization team is formally authorized, in writing, to use such names.
+   12. Names should not contain any unnecessary parts in the name, like "competition".
+   13. The competition nickname and ID must resemble the competition name.
+       1. The competition nickname and ID should only be modified if the competition name is too long (more than 32 characters). Exception: groups of competitions may have modified nicknames and IDs to ensure consistency with past and future competitions.
+    14. Abbreviations and acronyms should be avoided unless the competition name is too long (more than 32 characters).
+2. Location(s)
+   1. A location consists of the country, city, and venue.
+   2. The location should also include subregions of the country that are common to include with addresses (e.g. the state a U.S. city is located in).
+   3. Competitions will be accepted if there are no other competitions within 100 km driving distance and less than 19 days away.
```
Epride: I'm not sure this is the best wording for this. Just because the proximity policy is adhered to doesn't mean that the competition will necessarily be accepted.

documents/policies/Competition Requirements.md
cubewhiz:
```suggestion
### WCA Competition Requirements
```

documents/policies/Competition Requirements.md
cubewhiz:
```suggestion
   3. Competitions will be accepted regarding proximity if there are no other competitions within 100 km driving distance and less than 19 days away.
```

documents/policies/Competition Requirements.md
cubewhiz:
```suggestion
   1. The registration requirements must be clear and precise.
```

documents/policies/Competition Requirements.md
cubewhiz: This term is not new. These standards are also mentioned in the WQAC's Motion:
https://www.worldcubeassociation.org/documents/motions/10.2017.5%20-%20Quality%20Assurance%20Committee.pdf

I'm okay with keeping the term here and adding a definition to the Motions, but I don't think the definition belongs in this document.

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,150 @@
...
+      2. If used, the word "Championship" must be singular in the competition name.
+      3. Competitions which are not National, Continental, or World championships should avoid names that are similar to championship names.
+   3. Date(s)
+      1. National Championships should not be held on the same date(s) as Continental or World championships.
+      2. Continental Championships must not be held on the same date(s) as World Championships.
+   4. Championships should have all the officially recognized WCA events.
+   5. Championships must recognize and award champion titles and their respective prizes to the winners in the region.
+   6. Organization teams of National Championships should involve all the WCA Delegates of the respective countries, at the discretion of the Senior Delegate.
+10. Optional requirements
+    1. Require competitors to submit puzzles at a specific time before an event is scheduled to begin.
```
cubewhiz:
```suggestion
    1. The organization team may request to require competitors to submit puzzles at a specific time before an event is scheduled to begin.
```

documents/policies/Competition Requirements.md
cubewhiz:
```suggestion
    2. The organization team may request to prohibit competitors from participating in specific combinations of events.
```

cubewhiz: I think 10.1 and 10.2 should actually be nested underneath 10.3 and 10.1.1 and 10.1.2. Anyway, I've suggested as complete sentences.

documents/policies/Competition Requirements.md
ChrisWrightWCA: I agree that an addition to the definitions should cover this off.

documents/policies/Competition Requirements.md
ChrisWrightWCA: I agree it reads as though that is the deciding requirement.

ChrisWrightWCA: Much less ambiguous.

documents/policies/Competition Requirements.md
AlbertoPdRF: I agree that this is not the place for the definition, but I'm not sure if the Definitions Motion is either... given how such "quality standards" have evolved so much, especially in the last couple of years, I'm afraid we'll need to be updating that motion continuously :confused:

SAuroux: Hm, maybe this could be solved by requiring WQAC to come up with a "WCA Quality standards Policy" document. This might be overdue either way. 
The definition for "WCA quality standards" in the motions could be a cross-reference similar to the definition of "Regulations". At this point, it might also make sense to add a similar cross-reference defintion for "WCA competition requirements".

documents/policies/Competition Requirements.md
```diff
@@ -0,0 +1,156 @@
...
+      2. Different registration periods may have different registration requirements (e.g. restricting which events a competitor can register for).
+   3. Different locations for multi-location competitions may have different registration requirements (e.g. competitor limit).
+7. Events
+   1. Each event must have a planned number of rounds.
+      1. Each round of each event must have a planned format and time limit.
+      2. A round of an event must have planned advancement conditions if there is a subsequent round of the event.
+      3. A round may have a cutoff. A round with a cutoff is a combined round, with the formats as per [WCA Regulation 9b)](wca{regulations/#9b}).
+   2. Only results from events that appear on the competition website in the Events tab and Schedule will be recognized.
+      1. Other events may only be held unofficially.
+   3. If the main event of the competition is not the 3x3x3 Cube, it should be publicly mentioned in the competition's information.
```
SAuroux: I know that this is a little off-topic here, but motivated by the discussion about "WCA quality standards" below I want to remark that we should also add a defintion for "main event" at some point.

documents/policies/Competition Requirements.md
AlbertoPdRF: Yeah, that makes sense. We can discuss this on an email thread, it's time to merge this PR up! :slightly_smiling_face:

documents/policies/Competition Requirements.md
AlbertoPdRF: Maybe it could even fit into that "quality standards policy"? Anyway, let's discuss this elsewhere, the PR needs to be merged!

PR #49, opened 2019-01-13
WCA Data Protection Policy v1.0

AlbertoPdRF:


documents/policies/Data Protection.md
```diff
@@ -0,0 +1,26 @@
...
+2. The WCA Data Protection Committee has the following rights and duties:
+   1. Creation, implementation, and enforcement of all WCA Data Protection policies:
+      1. Researching regulations globally to ensure the WCA adopts the policy requirements to satisfy all global regulations.
+      2. Engaging with legal council on all matters relating to Data Protection with WCA Board approval.
+      3. Understanding the scope of which Data Protection policies are required and duly advising the WCA Board.
+   2. Acting as WCA liaison for Data Protection matters:
+      1. Managing requests from Registered Speedcubers and WCA Staff on Data Protection
+         1. Managing requests relating to data subject’s right to be forgotten.
+            1. Liaison with the WRT to request data subject data be removed or anonymized within the WCA Database.
+            2. Liaison with the WST to request that data subject data be removed or anonymized within the WCA Website.
```
jfly: nit "data subject data" might technically be the right term here, but it reads weird. I don't have a suggestion for a better phrase, sorry.

I'm not sure I understand the difference between "WCA Database" and "WCA Website" here. With very few hacky exceptions (like when we hardcoded the list of competitors with "In Memoriams"), ever bit of information like this *should* be in the WCA Database.

Long term, my vision would be for the WRT to handle all the removal/anonymization of data, but they would do it with tools written by the WST. In practice, right now I don't think it makes sense for the WST to invest a lot of time into building such tooling, when it's more practical to just go to the database and remove/scramble stuff by hand.

cubewhiz: How about "data subject's personal data" instead?

An example of something that was in the WCA website was a request to remove data from the old forums.

jfly: "data subject's personal data" sounds good to me.

> An example of something that was in the WCA website was a request to remove data from the old forums.

Forum data is in our database.

operge: I agree with the removal of line 22. Thanks Jeremy for the feedback!

ChrisWrightWCA: I too concur. Thanks for the feedback.

AlbertoPdRF: Change committed!

Issue #48, opened 2019-01-13
Remove Example Proximity Exception


Jambrose777:
Remove this line from the Competition Proximity Policy as it is not an (strong) argument that is acceptable for a proximity collision exception

> c. In the last 12 months, there have been fewer than 16 competitions in the area 100 km around the competition.

or change 16 to a more reasonable number like 8 (WCAT opinion).

AlbertoPdRF:
Ups, this should have been closed with #52!

PR #52, opened 2019-01-16
WCA Competition Proximity Policy v1.1

Jambrose777:
In the proximity policy, removed the argument for an exception of 16 nearby comps within the last year. Solves #48.

Of note: I was not able to successfully run the install and build scripts to ensure the PDF generated looks okay. Whenever the `sudo dpkg -i /tmp/pandoc.deb` command is run I get the error: `dpkg: error: failed to open package info file '/usr/local/var/lib/dpkg/status' for reading: No such file or directory`. It seems from here: https://github.com/Homebrew/legacy-homebrew/issues/18044 dpkg -i was removed? I haven't found much else on this.

### DIFF
--- a/documents/policies/Competition Proximity.md
+++ b/documents/policies/Competition Proximity.md
@@ -11,7 +11,6 @@ The purpose of this policy is to spread competitions in time and geographical ar
 2. Otherwise, the WCA Board should review the circumstances of the proposed competition. A Delegate requesting approval for such a competition must provide further information on why the competition should still be accepted. Possible arguments that support such a competition being accepted include:
     a. The two competitions don’t have events in common.
     b. The two competitions take place in different countries.
-    c. In the last 12 months, there have been fewer than 16 competitions in the area 100 km around the competition.
 
 ### 3x3x3 Fewest Moves Simultaneous Competitions
 A 3x3x3 Fewest Moves simultaneous competition is a WCA competition having only the 3x3x3 Fewest Moves event and taking place simultaneously in different locations.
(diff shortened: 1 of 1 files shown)


AlbertoPdRF:
Also, cc'ing @jonatanklosko to reply to:

> Of note: I was not able to successfully run the install and build scripts to ensure the PDF generated looks okay. Whenever the sudo dpkg -i /tmp/pandoc.deb command is run I get the error: dpkg: error: failed to open package info file '/usr/local/var/lib/dpkg/status' for reading: No such file or directory. It seems from here: Homebrew/legacy-homebrew#18044 dpkg -i was removed? I haven't found much else on this.

jonatanklosko:
@Jambrose777 you can install Pandoc manually (probably `brew install pandoc` will do), and then`wkhtmltopdf` as [here](https://github.com/thewca/wca-documents/blob/a6fdbd4dd24b129b899c80c81f1308e696b34bb5/bin/install_dependencies.sh#L14-L15). The `install_dependencies.sh` script targets the Travis environment, so it's not cross-platform.

Jambrose777:
> Thanks Jacob for sending in the PR! It's missing something, though: line 3 should be `### Version 1.1`.

Jambrose777:
Thanks @jfly  and @jonatanklosko  that does explain it.

Does the board not create the PDFs locally before merging in the PR?

cubewhiz:
@Jambrose777 The Board makes the relevant changes in a Google Doc so that we have an internal record of the version that we agreed upon. All policy updates require an official Board vote, though, which is in progress and must be completed before merging the PR. When the PR is merged, a new version of the policy will automatically be uploaded to the website. Does this answer your question?

PR #53, opened 2019-01-19
WCA Competition Name Policy v1.1

### LINKED ISSUE #50: Remove Silent Spring from the list of acceptable names
The WCAT believes that a competition name based solely on the competition events leaves the name too generic (although that in combination with location, for example, leaves a more unique name). This can lead to many competition in many locations being called the same thing. For instance a competition named: "Please Be Quiet 2019". Should we just allow this name to be taken from the first delegate who wants it? It's much better to add the location to the name in these cases.

Proposal, Remove:
> Silent Spring 2017 (competition is named after the events)

AlbertoPdRF:
Fixes #50.

### DIFF
--- a/documents/policies/Competition Name.md
+++ b/documents/policies/Competition Name.md
@@ -1,6 +1,6 @@
 # WCA Competition Name Policy
 <div class="version">
-### Version 1.0
+### Version 1.1
 </div>
 
 ## Purpose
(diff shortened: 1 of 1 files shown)


Jambrose777:
Should line 16 be modified in any way?

AlbertoPdRF:
I don't think that's necessary.

PR #60, opened 2019-02-24
[MotionAmendments-March2019] Finances.

AlbertoPdRF:
This PR is ready to review!

I had to add an extra space when nesting points to point 10 --  4 instead of the 3 necessary for the other points -- for the GitHub preview to show up correctly, although I'm not sure if when creating the PDF this will break anything. Any idea @jonatanklosko?


jonatanklosko:
@AlbertoPdRF I've just run the build locally and it looks good to me =)

AlbertoPdRF:
Thanks for confirming!

PR #62, opened 2019-02-24
Add Code of Ethics managed by WEC

jonatanklosko:
I believe that's all we need to let this file be built automatically.
Once there is an actual content we can modify the WCA `/documents` page to link to it.


jfly:
LGTM!

PR #103, opened 2019-08-23
Remove Dues for Multi-country FMC competitions

cubewhiz:
Updates to v1.3

### DIFF
--- a/documents/policies/external/Dues System.md
+++ b/documents/policies/external/Dues System.md
@@ -1,10 +1,10 @@
 # WCA Dues System Policy
 <div class="version">
-### Version 1.2
+### Version 1.3
 </div>
 
 ## Purpose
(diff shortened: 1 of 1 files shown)


documents/policies/external/Dues System.md
```diff
@@ -35,32 +35,17 @@ per-competitor-per-competition basis.
 To calculate the Due amount owed for a competition, the following procedure will be used:
 
 1. The country used to determine the Due amount owed is the country in which the competition takes place.
-   1. For 3x3x3 Fewest Moves simultaneous competitions, please see Appendix B for how to calculate the Due amount owed.
+   1. For 3x3x3 Fewest Moves simultaneous competitions that take place in more than one country, the Due amount will be waived.
```
Jambrose777: I'm sure the board is already aware of these, but felt I should comment:

- Brazil has FMC competitions (last one had 6 locations): https://www.worldcubeassociation.org/search?q=brazil%20fmc
- USA used to and I guess there is potential for them in the future, which have many locations: https://www.worldcubeassociation.org/search?q=FMC%20USA

These would be excluded from the exception.

cubewhiz: That's correct--the exception will only apply to FMC competitions that take place in multiple countries because of the difficulty in invoicing those. FMC competitions in a single countries are easily billable.

PR #110, opened 2019-10-08
Code of Ethics update

c-goodyear:
Version 1.0 to version 2.0 of the Code of Ethics after a WEC meeting and subsequent Board, Senior Delegate, and Team/Committee Leader input.

### DIFF
--- a/documents/Code of Ethics.md
+++ b/documents/Code of Ethics.md
@@ -1,87 +1,97 @@
 # Code of Ethics
 <div class="version">
-### Version 1.0
+### Version 2.0
 </div>
 
-The Code of Ethics must be considered by all Staff members of the World Cube Association (WCA), referred to in this document as “WCA Staff”, as a standard for actions and decisions they will be making when acting on behalf of the WCA within their role. “Listed Delegates” refers to WCA Delegates who are publically displayed on the “General info” tab of an official WCA Competition.
+The Code of Ethics must be considered by all Staff members of the World Cube Association (WCA), referred to in this document as “WCA Staff”, as a standard for actions and decisions they will be making when acting on behalf of the WCA within their role. “Listed Delegates” refers to WCA Delegates who are publicly displayed on the “General info” tab of an official WCA Competition.
 
 These actions and decisions should be made in such a manner that, if consistently performed by WCA Staff, the [Spirit and Mission](wca{documents/motions/01.2019.1 - Spirit.pdf}) of the WCA would continue to be upheld while not breaking any [Regulations](wca{regulations}), [Motions, or Policies](wca{documents}).
 
-This Code covers previously known issues, as well as issues that the WCA Ethics Committee (WEC) foresees as requiring additional guidance. This document will be updated periodically in order to incorporate new topics as they are brought to the attention of the WEC. This document will also be reviewed by the WEC regularly and will be updated as required.
+This Code covers previously known issues, as well as issues that the WCA Ethics Committee (WEC) foresees as requiring additional guidance. This document will be updated periodically to incorporate new topics as they are brought to the attention of the WEC, and will also be reviewed regularly.
 
 The official communication channel for the WEC is via email. For any concerns about the ethical conduct of a WCA Staff member, contact the WEC: [ethics@worldcubeassociation.org](mailto:ethics@worldcubeassociation.org). If there are any concerns about the ethical conduct of the WCA Ethics Committee, either the Committee itself or an individual member, please contact the WCA Disciplinary Committee (WDC) at: [disciplinary@worldcubeassociation.org](mailto:disciplinary@worldcubeassociation.org).
 
 <div class="page-break"></div>
 
-1. **Dealing with Finances**
-   1. WCA Staff must not use money obtained from their role within the WCA, including but not limited to competitor fees, expense reimbursement, equipment policies, and sponsorships, for personal profit.
-   2. Competition reimbursements for Delegates should only include costs incurred as a direct result of attending the competition as a listed Delegate.
-      1. This may include the following personal expenses:
+1. **Finances**
+   1. WCA Staff must not use money, discounts, or goods obtained from their role within the WCA for personal profit.
+      1. Examples include:
+         1. Competitor fees
+         2. Equipment
+         3. Expense reimbursement
+         4. Sponsorships
+   2. Reimbursements for Delegates must only include costs incurred as a direct result of attending the competition as a listed Delegate.
+      1. This may include the following expenses:
          1. Transportation
          2. Accommodation
          3. Equipment and Printing
-      2. Deviations from 1.2.1 require WEC approval before the expense is incurred.
-      3. When Delegates are not officially listed and are just attending a competition as competitors, their expenses, including registration fees, are not valid for reimbursement.
-   3. Registration fees for listed Delegates and Organizers are permitted to be waived. Registration fees should only be waived for competitors if they perform a role necessary for the running of the competition.
-      1. Sole reasons for waiving a competitors registration fee must not include:
+      2. Expenses not laid out in 1.2.1 require WEC approval, and WFC if necessary.
+   3. Registration fees for listed Delegates and Organizers may be waived. Registration fees should only be waived for competitors if they perform a role significant to the running of the competition.
+      1. Waiving of registration fees must not be done for the sole reason of people being the following (or similar):
          1. Family
          2. Friends
-      2. Reminder: Regulations [1e2)](wca{regulations/#1e2}) and [1f2)](wca{regulations/#1f2}) should be used to help  fill scrambling and judging roles during a competition.
-      3. Competition volunteer application forms should be available for competitions to help manage and standardize who will be waived for competition volunteer purposes.
-      4. If the competition venue is provided for free or for a discounted fee, members of a relevant group can be partially or fully waived from paying the competition’s registration fee if this is a requirement of the venue hire.
-         1. Delegates should be wary that any such deal should still be fair and avoid bias towards any unreasonably formed group.
-   4. Listed Delegates must request and review an outlined budget for all of their competitions before announcement. Any concerns in the competition budget should be discussed with an organizer and/or the people in charge of the competition finances. Delegates and organizers should work together to find a resolution to any concern.
-      1. Competition budgets must be held by the Delegate and if necessary can be requested for review by the WCA Board, WEC, WFC or the Senior Delegate.
+      2. Reminder: Regulations [1e2)](wca{regulations/#1e2}) and [1f2)](wca{regulations/#1f2}) should be used to help fill scrambling and judging roles during a competition.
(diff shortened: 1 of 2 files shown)


documents/Code of Ethics.md
```diff
@@ -1,87 +1,97 @@
...
-      1. Wearing the apparel of a sponsor while acting within their WCA Staff role at a competition
-      2. Promotions in emails to competitors
-      3. References in the email signature block
-   4. WCA Delegates must not wear sponsorship apparel of a personal sponsor during a competition in which they are listed as a Delegate.
-      1. Exception: Listed Delegates may only wear their sponsorship apparel while they are competing at a competition.
+      1. Any sponsorship details of a WCA Staff member as a competitor must be reported to the WEC.
+   3. WCA Staff must not use their role to promote a third party brand who are not [a WCA-recognized regional organization](https://www.worldcubeassociation.org/organizations), the competition sponsor(s), or sponsors of the WCA. This includes but is not limited to:
+      1. Promotions in emails to competitors
+      2. References in the email signature block
+   4. WCA Delegates must not wear apparel which could reasonably be interpreted as an official affiliation or agreement with the WCA.
```
Jambrose777: Is it possible to make this more specific? I feel like this would prevent unlisted Delegates at a competition like Worlds from wearing what they wanted at any time.

documents/Code of Ethics.md
```diff
@@ -1,87 +1,97 @@
...
          3. Event frequency in the area
          4. Inexperience of the community
-         5. Inexperience of the organizer
-         6. Past problems with an event held by the organizer/community
-      2. Delegates should work alongside the organizer(s) to plan and resolve any of these reasons listed under 3.3.1.
-   4. Delegates and Organizers should only be officially listed for a competition if they perform a necessary role in the running of the event.
+         5. Inexperience of the Organizer
+         6. Past problems with an event held by the Organizer/community
+      2. Delegates should work alongside the Organizer(s) to plan and resolve any of these reasons listed under 3.3.1.
+   4. Delegates and Organizers should only be officially listed for a competition if they perform a significant role.
```
EdHollingdale: What is a "significant role" - I feel it should be defined? I would list a Delegate if they are at the competition and handle a minor incident even to the trivial nature of "yes, that is a DNF". Obviously per the regulations even the granting of a basic extra attempt must be done by a delegate, as with handling of scrambles, which would necessitate listing the delegate, even if they did not meet the "significant role" barrier.

documents/Code of Ethics.md
```diff
@@ -1,87 +1,97 @@
...
    2. WCA Staff must not receive sponsorship(s) for their role within the WCA.
-      1. Any sponsorship details of a WCA Staff member must be reported to the WEC.
-   3. WCA Staff must not promote a third party brand who are not the competition sponsor(s) or sponsors of the WCA, within their role in the WCA. This includes but is not limited to:
-      1. Wearing the apparel of a sponsor while acting within their WCA Staff role at a competition
-      2. Promotions in emails to competitors
-      3. References in the email signature block
-   4. WCA Delegates must not wear sponsorship apparel of a personal sponsor during a competition in which they are listed as a Delegate.
-      1. Exception: Listed Delegates may only wear their sponsorship apparel while they are competing at a competition.
+      1. Any sponsorship details of a WCA Staff member as a competitor must be reported to the WEC.
+   3. WCA Staff must not use their role to promote a third party brand who are not [a WCA-recognized regional organization](https://www.worldcubeassociation.org/organizations), the competition sponsor(s), or sponsors of the WCA. This includes but is not limited to:
```
EdHollingdale: What about a not officially recognised "sub regional organisation"? I don't feel these should be excluded, and they should be encouraged, since they enable better management of competition funds in a manner compliant with the code of ethics!

documents/Code of Ethics.md
c-goodyear: It was changed to be worded in a way that moves more towards an explanation of what we are trying to acheive with it being in the CoE. As opposed to nailing down every specific instance like the regulations, which would be impossible for us to do anyway because of WEC's scope.

This would prevent unlisted Delegates from wearing sponsored garments while not competing, correct. However this closes the loophole you pointed out before about unlisteds becoming listed causing issues if they were promoting a third party during that competition. That instance can occur at more competitions than the biennial World or Continental championships, so on balance we decided that the same rule for all people holding a Delegate position was fairer.

documents/Code of Ethics.md
```diff
@@ -1,87 +1,97 @@
...
       1. All official WCA email communication must utilize the @worldcubeassociation.org addresses.
       2. WCA Staff must be aware that emails sent from @worldcubeassociation.org email addresses to external parties may be perceived as being on behalf of the WCA. Personal email addresses should be used to avoid this perception.
          1. Example: Contacting sponsors for a competition should use personal email addresses.
-   3. WCA Delegates must communicate with other Delegates in their region and/or in any region they are planning to Delegate a competition.
-   4. All work done within the WCA is confidential and must not be disclosed to non-WCA Staff members without explicit permission from relevant members.
+   2. WCA Delegates must communicate with other Delegates in their region and/or in any region they are planning to delegate a competition.
+   3. All work done within the WCA is confidential and must not be disclosed to non-WCA Staff members without explicit permission from relevant members.
+      1. Delegates may provide relevant feedback to the Organizers but must not give visibility of incidents, or sections written by other Delegates.
```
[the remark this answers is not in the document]
c-goodyear: I'm blanking on a re-word for that section but I don't infer the same as you, this allows copy/pasting of sections such as "Organization" instead of the Delegate having to re-write something if they want to give feedback to the Organizer, reducing their workload if this is requested.

I don't understand why we need to make an exception for other Staff members seeing reports when the phrasing is "must not be disclosed to non-WCA Staff members".

As for training Delegates, this definitely seems like something WQAC would be able to produce. A kind of dummy report with mock incidents etc. For incident context and reasoning, surely the best time to do that would be at the competition when the incident has just occured.

documents/Code of Ethics.md
c-goodyear: If they are not recognized then they haven't been "vetted" by the WCA so we have no basis for what standards they hold or what their Mission is etc. I also think it's naive that you feel there are no regional organizations who arrange their finances in a way that would not be compliant with the CoE.

Don't forget this covers all Staff, for example if a communication team member started making WCA posts to promote a non-recognized regional organization, I think you can agree it would be an inappropriate use of their role and access.

documents/Code of Ethics.md
c-goodyear: Some people had issues with the word "necessary" in the previous version. Granting an extra seems significant to me, given that you have to be a WCA Delegate in order to grant one. (I.e. if they require listing because of an action they took -> that action was significant)

There are a million and one things that can happen at competitions to Delegates, I don't think that turning the Code of Ethics into the Regulations where we try and cover every possible scenario is feasible.

documents/Code of Ethics.md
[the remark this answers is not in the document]
EdHollingdale: I agree that having a dummy report produced by WQAC would be useful. However I see no reason why a typical incidents section of a delegate report cannot be shared to organisers (with the removal of specific names, or sensitive incidents, as required). It is a clear way of showing what incidents are/aren't worth highlighting, in which circumstances to seek WRC feedback, and how to clearly articulate an incident and the outcome.

I had missed that this was a subpoint to something that clearly stated "non-WCA Staff". I agree no exception then is required then.

documents/Code of Ethics.md
EdHollingdale: If we are not being prescriptive about what "significant" is then I don't see the point having it there. I actually thing a broader problem is Delegates _not_ being listed even when they were at the competition and staffing/helping out such they were perceived as listed delegates (even if they never granted an extra).

documents/Code of Ethics.md
EdHollingdale: I certainly don't think that just because it is an organisation it is compliant with the CoE, but I think that having things structured via a regional organisation is better than not doing so.

documents/Code of Ethics.md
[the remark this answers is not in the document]
c-goodyear: 1. Delegates may provide relevant feedback to the Organizers but must not give visibility of **sensitive information**, or sections written by other Delegates.

Thoughts on that phrasing?

documents/Code of Ethics.md
```diff
@@ -1,87 +1,97 @@
...
 3. **Conduct of Official WCA Competitions**
-   1. Decisions on incidents required to be made by a WCA Delegate must only be made by those officially listed on the competition page.
+   1. Decisions on incidents required to be made by a WCA Delegate must only be made by those listed on the WCA competition page.
```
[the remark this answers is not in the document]
c-goodyear: If we allow Delegates to act in their Staff capacity when they are not listed as such publicly, then it causes other problems. Competitors (and their parents, in the case of minors) can easily see who they should be asking to talk to by checking the website if an incident occurs where the Delegate needs to be informed or consulted. We've had to deal with issues in the past caused by Delegates making calls when they were not listed. This allows everyone to know who should be making decisions.

documents/Code of Ethics.md
c-goodyear: I can't imagine how we can cover all cases Delegates can encounter and deal with, therefore wording such as "significant" (previously "necessary") is required to avoid attempting to write an endless list which would still leave holes. Granted, some perceptions of "significant" may be different than others, but we already see this in the regulations where being prescriptive is the name of the game.

c-goodyear:
Not sure what happened but my small WCRP fix also got appended to this PR...

documents/Code of Ethics.md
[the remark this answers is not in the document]
EdHollingdale: I much prefer this phrasing. I think it gives the flexibility to delegates to help organisers, while keeping the correct information around privacy etc.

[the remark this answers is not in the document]
c-goodyear:
```suggestion
      1. Delegates may provide relevant feedback to the Organizers but must not give visibility of sensitive information, or sections written by other Delegates.
```

documents/Code of Ethics.md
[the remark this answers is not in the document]
EdHollingdale: Yes, on second thought I agree. We should be forcing Delegates to plan ahead speak to any other Delegates who may be attending to set out clearer roles before the competition, and speak to WCAT ahead of time to list them if appropriate. Recent experiences have taught me that setting clear expectations for organisers and Delegates before a competition is best for all concerned.

documents/Code of Ethics.md
EdHollingdale: I see your point, and I think "significant" is better wording than "necessary". I think we should make it clear when messaging these changes to Delegates that any task defined in the regulations as a task for the Delegate meets the criteria of significant.

c-goodyear: That's fair enough, I will outline that in the announcement to Staff when it's sent! :-)

Issue #111, opened 2019-10-28
WCRP Update Umbrella October 2019


DanielEgdal:
WCAT has discussed a series of changes. A thread will be opened for all points so we can fully discuss them. For WCA Staff, you can have a look at the document we used for internal discussion: https://docs.google.com/document/d/18q_xv86_cQynytiLynL4_hhqr2g-jgEYtXXvxiXPdhM/edit?usp=sharing

- [x] Universal registration requirements. #112 
- [x] Event additions. #113
- [x] Main event. #114
- [x] Scorecards. #115 
- [x] Grammar update of 8.2.2.2.1. (looks to be solved together with https://github.com/thewca/wca-documents/pull/110)

These are necessesary to be addressed as well, but WCAT skipped these when we initially drafted the above points:
- [x] Dealing with proceeds going to charities. #108 
- [x] Clarify competitor limits per event. #104 
- [x] Clarify competitor limits per location for FMC Simultaneous competitions. #116 
- [x] Clarify waiting lists. https://github.com/thewca/wca-documents/issues/118
- [x] Clarify deadlines for updating registrations https://github.com/thewca/wca-documents/issues/119

cubewhiz:
Thanks for compiling this list. I've given suggestions in individual issues. Please submit the PR for review when you have the availability.

DanielEgdal:
Thanks Bob! I have created a compiled PR of the issues (with the exception of the Scorecards issues). https://github.com/thewca/wca-documents/pull/117

DanielEgdal:
Two more points were added.

PR #117, opened 2019-11-03
WCRP Update Version 2.0

DanielEgdal:
Fixes https://github.com/thewca/wca-documents/issues/111.

### DIFF
--- a/documents/policies/external/Competition Requirements.md
+++ b/documents/policies/external/Competition Requirements.md
@@ -42,6 +42,7 @@ competition.
       2. The competition website cannot advertise any specific products or services of the sponsor, including any endorsements of quality or price.
    3. The competition website may acknowledge the sponsorship contribution.
 10. Competitions with proceeds going to a charity must indicate so on the competition website, preferably in the Information section of the 'General info' tab. The name of the charity must be included. Any such charity needs to be recognized in the country where the competition is held.
+   1. Having this indicated on the competition website must be done as soon as the charity of the proceeds for has been decided. This must be done prior to the competition taking place. Decisions of charities decided after the competition has taken place must be approved by the Ethics Committee.
 
 If WCA Regulations are not correctly observed during a competition, results may be modified at the discretion of the WCA Delegate, WCA Regulations Committee, WCA Results Team, WCA Disciplinary Committee, or the WCA Board.
 
@@ -116,19 +117,20 @@ The WCA Delegate must submit the following information when requesting approval 
             2. Not having enough money to pay the registration fee.
    2. The registration period(s) must be a clearly defined time range with a start and end time and date.
       1. Each registration period must be announced publicly on the WCA website at least 48 hours before it starts.
-         1. The competition's WCA Delegate(s) and/or organizer(s) may register for the competition prior to the first registration period, however their registration(s) must not be approved until the first registration period has started.
+         1. An active registration period may be extended after request of the Delegate.
+         2. The competition's WCA Delegate(s) and/or organizer(s) may register for the competition prior to the first registration period, however their registration(s) must not be approved until the first registration period has started.
       2. Different registration periods may have different registration requirements (e.g. restricting which events a competitor can register for).
    3. The deadline for updating registered events for competitors with completed registrations must be outlined.
-      1. This deadline will either be at the same time as registration closes or after registration has closed.
+      1. This deadline must not be earlier than the time registration is set to close. 
          1. The deadline may be set to be just before an event is scheduled to take place. 
       2. Competitors may choose not to compete in an event they are registered for.
-   4. The organization team must outline the conditions of the waiting list, and if one is in place.
+   4. The organization team must outline the conditions of the waiting list.
       1. If a waiting list is in place:
          1. Registrants will be placed on the waiting list if they have completed all requirements for a valid registration, but the competitor limit has been filled. Registrants from the waiting list will be accepted in the order that their registration was completed once a spot becomes available.
          2. It must be stated what the deadline is to be accepted for a registrant on the waiting list.
             1. This deadline should be the same as the deadline for refunds for competitors canceling their registration, if (partial) refunds are to be given.
-         3. The organization team may specify a limit to the waiting list. Once the limit of the waiting list has been reached, the Delegate should inform the WCAT to close the registration period.
-      2. If a waiting list is not in place, the Delegate should inform the WCAT once the competitor limit has been filled so the registration period can be closed.
+         3. The organization team may specify a limit to the waiting list. Once the limit of the waiting list has been reached, the Delegate should inform the WCAT to close the registration period. 
+            1. In case a limit was not set in advance, the Delegate may contact the WCAT to request to have the registration period closed if the waiting list has become unreasonably long.
    5. Different locations for multi-location competitions may have different registration requirements (e.g. competitor limit).
 7. Events
    1. Each event must have a planned number of rounds.
@@ -137,8 +139,8 @@ The WCA Delegate must submit the following information when requesting approval 
       3. A round may have a cutoff. A round with a cutoff is a combined round, with the formats as per [WCA Regulation 9b)](wca{regulations/#9b}).
    2. Only results from events that appear on the competition website in the Events tab and Schedule will be recognized.
       1. Other events may only be held unofficially.
-   3. If the main event of the competition is not the 3x3x3 Cube, it should be publicly mentioned in the competition's information.
-      1. The main event should be the main focus of the competition by being recognizable in the competition schedule and having an appropriate proportion of rounds and attempts.
+   3. If the main event of the competition is not the 3x3x3 Cube, it should be publicly mentioned in the competition's information section of the 'General info' tab.
+      1. The main event should be recognizable in the competition schedule and having an appropriate proportion of rounds and attempts.
    4. An additional event may be added upon the WCA Delegate's request at the discretion of the WCAT. The competition must meet the following requirements for an additional event to be added:
       1. The competition is at least 7 days away.
       2. A registration period for the competition must have a minimum of 72 hours before it closes.
--- a/documents/policies/external/Competition Requirements.md
+++ b/documents/policies/external/Competition Requirements.md
@@ -19,9 +19,7 @@ An official WCA Competition must:
 6. Have at least one official event.
 7. Have at least one competitor.
 8. Have at least one designated WCA Delegate in attendance.
-   1. Exception: in case the WCA Delegate cannot attend the competition, the WCA Board may
-appoint a trusted person to act as a temporary WCA Candidate Delegate for the
-competition.
+   1. Exception: in case the WCA Delegate cannot attend the competition, the WCA Board may appoint a trusted person to act as a temporary WCA Delegate for the competition.
    2. The WCA Delegate for a competition is responsible for:
       1. Submitting the Delegate Report to the WCA Regulations Committee and the WCA Quality Assurance Committee regarding adherence to WCA Regulations and quality standards during the competition, the overall course of the competition, and any incidents that happened. The report must be submitted within one week of the end of the competition.
       2. Submitting the competition results to the WCA Results Team within one week of the end of the competition.
--- a/documents/policies/external/Competition Requirements.md
+++ b/documents/policies/external/Competition Requirements.md
@@ -141,6 +141,8 @@ The WCA Delegate must submit the following information when requesting approval 
       1. Other events may only be held unofficially.
    3. If the main event of the competition is not the 3x3x3 Cube, it should be publicly mentioned in the competition's information section of the 'General info' tab.
       1. The main event should be recognizable in the competition schedule and having an appropriate proportion of rounds and attempts.
+      2. Only WCA sanctioned events may be considered the main event of the competition.
+      3. There may be no main event of the competition.
    4. An additional event may be added upon the WCA Delegate's request at the discretion of the WCAT. The competition must meet the following requirements for an additional event to be added:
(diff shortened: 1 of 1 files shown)


DanielEgdal:
Two more points (#118 & #119 ) need discussion before they are added to this PR.

DanielEgdal:
The two points were added.

documents/policies/external/Competition Requirements.md
```diff
@@ -97,20 +98,38 @@ The WCA Delegate must submit the following information when requesting approval
...
+      1. This deadline will either be at the same time as registration closes or after registration has closed.
+         1. The deadline may be set to be just before an event is scheduled to take place. 
+      2. Competitors may choose not to compete in an event they are registered for.
+   4. The organization team must outline the conditions of the waiting list, and if one is in place.
+      1. If a waiting list is in place:
+         1. Registrants will be placed on the waiting list if they have completed all requirements for a valid registration, but the competitor limit has been filled. Registrants from the waiting list will be accepted in the order that their registration was completed once a spot becomes available.
+         2. It must be stated what the deadline is to be accepted for a registrant on the waiting list.
+            1. This deadline should be the same as the deadline for refunds for competitors canceling their registration, if (partial) refunds are to be given.
+         3. The organization team may specify a limit to the waiting list. Once the limit of the waiting list has been reached, the Delegate should inform the WCAT to close the registration period.
```
EdHollingdale: I like this. Does this limit have to be specified in advance, or will WCAT allow closing registrations once it gets to a high number, as they have in the past? Either way, it would be good to specify in the policy I feel.

documents/policies/external/Competition Requirements.md
```diff
@@ -119,6 +138,11 @@ The WCA Delegate must submit the following information when requesting approval
    2. Only results from events that appear on the competition website in the Events tab and Schedule will be recognized.
       1. Other events may only be held unofficially.
    3. If the main event of the competition is not the 3x3x3 Cube, it should be publicly mentioned in the competition's information.
+      1. The main event should be the main focus of the competition by being recognizable in the competition schedule and having an appropriate proportion of rounds and attempts.
```
EdHollingdale: What does this mean? I would argue that the "main focus" of a competition is having fun and meeting friends. As an organiser I normally announce the winner of the main event as the winner of the competition and make a point of it at awards, which I feel is sufficient but I'm not sure would meet this threshold.

At "standard" competitions 3x3 is just the last event held with the most competitors and most rounds - it could be a main/not a focus/one of many focuses depending on the person. As a different example, for a competition the "main focus" could be "side-events" but you pick e.g. Skewb because it has the last final held and has the equal most rounds, but that doesn't make it the "main focus", although it could be the "main event". For "PBQ" competitions FMC could be the main event, but it is hard make it recognisable given the nature of the event and competition logistics requirements, however it can be made worthy of being a main event during the presentation of awards/prizes, for example.

documents/policies/external/Competition Requirements.md
```diff
@@ -119,6 +138,11 @@ The WCA Delegate must submit the following information when requesting approval
...
+   4. An additional event may be added upon the WCA Delegate's request at the discretion of the WCAT. The competition must meet the following requirements for an additional event to be added:
+      1. The competition is at least 7 days away.
+      2. A registration period for the competition must have a minimum of 72 hours before it closes.
```
EdHollingdale: If the competition is e.g. 10 days away but registration is open for only another 48 hours would it be possible to extend it? It would be good to note this, e.g. "The registration period may be extended if it is still open if this is not the case" or "It is not possible to extend registration in order to enable another event to be added".

documents/policies/external/Competition Requirements.md
```diff
@@ -41,6 +41,7 @@ competition.
       1. The competition website cannot link to web pages that advertise specific products or services of the sponsor (unless it is the sponsor’s homepage).
       2. The competition website cannot advertise any specific products or services of the sponsor, including any endorsements of quality or price.
    3. The competition website may acknowledge the sponsorship contribution.
+10. Competitions with proceeds going to a charity must indicate so on the competition website, preferably in the Information section of the 'General info' tab. The name of the charity must be included. Any such charity needs to be recognized in the country where the competition is held.
```
SAuroux: When does this need to be done? Before announcement? Before the competition takes place? What if organizers realize they have an unexpected excess, can they still decide to give this to a charity after the competition has taken place? If yes, what is the remaining purpose of this paragraph?

documents/policies/external/Competition Requirements.md
```diff
@@ -97,20 +98,38 @@ The WCA Delegate must submit the following information when requesting approval
...
+         2. Exceptions must not be based on:
+            1. Missing the registration deadline.
+            2. Not having enough money to pay the registration fee.
    2. The registration period(s) must be a clearly defined time range with a start and end time and date.
       1. Each registration period must be announced publicly on the WCA website at least 48 hours before it starts.
          1. The competition's WCA Delegate(s) and/or organizer(s) may register for the competition prior to the first registration period, however their registration(s) must not be approved until the first registration period has started.
       2. Different registration periods may have different registration requirements (e.g. restricting which events a competitor can register for).
-   3. Different locations for multi-location competitions may have different registration requirements (e.g. competitor limit).
+   3. The deadline for updating registered events for competitors with completed registrations must be outlined.
+      1. This deadline will either be at the same time as registration closes or after registration has closed.
```
SAuroux: Better: This deadline must not be earlier than the time as registration closes.

documents/policies/external/Competition Requirements.md
```diff
@@ -97,20 +98,38 @@ The WCA Delegate must submit the following information when requesting approval
...
+         1. The deadline may be set to be just before an event is scheduled to take place.
```
SAuroux: I would rather propose to add this to the top-level paragraph instead as default if no other point in time specified.

documents/policies/external/Competition Requirements.md
```diff
@@ -97,20 +98,38 @@ The WCA Delegate must submit the following information when requesting approval
...
+            1. Having an unsupported credit card for paying the registration fee.
+         2. Exceptions must not be based on:
+            1. Missing the registration deadline.
+            2. Not having enough money to pay the registration fee.
    2. The registration period(s) must be a clearly defined time range with a start and end time and date.
       1. Each registration period must be announced publicly on the WCA website at least 48 hours before it starts.
          1. The competition's WCA Delegate(s) and/or organizer(s) may register for the competition prior to the first registration period, however their registration(s) must not be approved until the first registration period has started.
       2. Different registration periods may have different registration requirements (e.g. restricting which events a competitor can register for).
-   3. Different locations for multi-location competitions may have different registration requirements (e.g. competitor limit).
+   3. The deadline for updating registered events for competitors with completed registrations must be outlined.
```
SAuroux: Will this be another timestamp to be set on the WCA website?

documents/policies/external/Competition Requirements.md
```diff
@@ -97,20 +98,38 @@ The WCA Delegate must submit the following information when requesting approval
...
+      1. This deadline will either be at the same time as registration closes or after registration has closed.
+         1. The deadline may be set to be just before an event is scheduled to take place. 
+      2. Competitors may choose not to compete in an event they are registered for.
+   4. The organization team must outline the conditions of the waiting list, and if one is in place.
+      1. If a waiting list is in place:
+         1. Registrants will be placed on the waiting list if they have completed all requirements for a valid registration, but the competitor limit has been filled. Registrants from the waiting list will be accepted in the order that their registration was completed once a spot becomes available.
+         2. It must be stated what the deadline is to be accepted for a registrant on the waiting list.
```
SAuroux: Will this be another timestamp to be set on the WCA website?

documents/policies/external/Competition Requirements.md
```diff
@@ -97,20 +98,38 @@ The WCA Delegate must submit the following information when requesting approval
...
+            1. This deadline should be the same as the deadline for refunds for competitors canceling their registration, if (partial) refunds are to be given.
+         3. The organization team may specify a limit to the waiting list. Once the limit of the waiting list has been reached, the Delegate should inform the WCAT to close the registration period.
+      2. If a waiting list is not in place, the Delegate should inform the WCAT once the competitor limit has been filled so the registration period can be closed.
```
SAuroux: I agree with the thought behind this, but the implementation is awful. You should not require from delegates to manually notify WCAT at a certain point in time when this could just as well be managed by a software solution that automatically prevents registration once the limit is reached.

documents/policies/external/Competition Requirements.md
SAuroux: I agree with the thought behind this, but the implementation is awful. You should not require from delegates to manually notify WCAT at a certain point in time when this could just as well be managed by a software solution that automatically prevents registration once the limit is reached.

documents/policies/external/Competition Requirements.md
SAuroux: Furthermore: What happens when there is no waiting list in place, the limit has been reached once, but there are cancelations hereafter? Does this need to be solved with a second registration period?

documents/policies/external/Competition Requirements.md
SAuroux: I agree with Ed. I would propose the following instead:

The main event should be ~~the main focus of the competition by being~~ recognizable in the competition schedule and having an appropriate proportion of rounds and attempts.

documents/policies/external/Competition Requirements.md
DanielEgdal: Personally, I see it as being the default that people cannot change their registration after registration closes due to the implications it may have.  That is why I added it as a clarifying point.

documents/policies/external/Competition Requirements.md
DanielEgdal: Personally, I think there is really no reason why not to have a waiting list now that I am thinking about it. 
There have been several issues on github beforehand that it should close automatically but it hasn't been implemented yet. So for the time being I believe WCAT should just do the manual work.

documents/policies/external/Competition Requirements.md
DanielEgdal: Personally, advancements and cutoffs play a big part in how I see a main event as well. So what Sébastien wrote seems good to me.

documents/policies/external/Competition Requirements.md
DanielEgdal: I tried to clarify this earlier in the WCRP that registration periods may be extended. Like Clement mentioned, the WCAT would not deny extending an active registration period so this may be done in order to fulfill the requirements of adding an event.

documents/policies/external/Competition Requirements.md
DanielEgdal: I have currently mentioned this must be done in advance and if done afterwards, WEC must approve it. I will discuss this with WEC about the liability of this.

c-goodyear: @pokekrom the problem with adding such things _after_ people paid their registration is that they may not want their money going towards that particular charity, and they now have no choice (especially if there is no refund).

SAuroux: @c-goodyear Is that a problem though? I would rather argue that once paid, the registration fee is no longer _their money_. It should be sufficient if WEC (or more generally WCA) argrees with a certain charity being a credible choice to send competition excess to.

documents/policies/external/Competition Requirements.md
SAuroux: Ok, let's see how this works out in practice.

documents/policies/external/Competition Requirements.md
c-goodyear: @SAuroux Taking someone's money and then changing what their money is going towards is pretty shifty, especially since I can't think of any charities that are _directly_ in alignment with the WCA's mission/spirit. That's why it's not an issue with funds going towards national/local organizations where the funds will directly assist further WCA competitions.

documents/policies/external/Competition Requirements.md
```diff
@@ -97,28 +99,52 @@ The WCA Delegate must submit the following information when requesting approval
...
 7. Events
    1. Each event must have a planned number of rounds.
       1. Each round of each event must have a planned format and time limit.
       2. A round of an event must have planned advancement conditions if there is a subsequent round of the event.
       3. A round may have a cutoff. A round with a cutoff is a combined round, with the formats as per [WCA Regulation 9b)](wca{regulations/#9b}).
    2. Only results from events that appear on the competition website in the Events tab and Schedule will be recognized.
       1. Other events may only be held unofficially.
-   3. If the main event of the competition is not the 3x3x3 Cube, it should be publicly mentioned in the competition's information.
+   3. If the main event of the competition is not the 3x3x3 Cube, it should be publicly mentioned in the competition's information section of the 'General info' tab.
+      1. The main event should be recognizable in the competition schedule and having an appropriate proportion of rounds and attempts.
```
Ivan-Brigidano: Would it be a requirement that the main event is an official WCA sanctioned event?

DanielEgdal: It isn't possible to have an unofficial event as the main event as the WRT would then not have anything to make a post about when announcing the winner. 
I realise now that this is implying that there needs to be a main event, which isn't the case (i.e. there may be no main event).

Ivan-Brigidano: Thanks for the clarification. I would suggest the WCAT to explicitly say that only official events are eligible to be the main event, on the WCRP. I can easily see organizers trying to have a [SEE event](https://speedcubingextraevents.org) as the main event.

DanielEgdal: I clarified this now.

PR #120, opened 2019-11-10
Integrate wca-edudoc style into compiler toolchain

suushiemaniac:
Porting of the format used for https://github.com/suushiemaniac/wca-edudoc into the official documents repo.

Basically introduces a second loop that uses a different CSS stylesheet for `wkhtml` as well as custom headers and footers. The principle architecture has been discussed with and approved by @viroulep.

In particular, this PR will slightly change the structure of the `build/` folder: Instead of directly replicating the contents _inside_ the `documents/` folder it instead copies the folder _as a whole_ so that we will now have:
```
build/
-- documents/
---- Code of Ethics.pdf
---- etc.
-- edudoc/
```

This change is for the website to be able to differentiate which documents to link on which page (in an upcoming PR)

I have also included the current working state of the competitor tutorial at my `wca-edudoc` repo as a sanity check.


bin/build.sh
```diff
@@ -2,17 +2,46 @@
 
 wca_url="https://worldcubeassociation.org/"
 
-# Find all Markdown files and build PDFs out of them.
+# Find formal and legal Markdown files and build PDFs out of them.
 find 'documents' -name '*.md' | while read file; do
+  echo "Processing $file, vanilla style"
```
suushiemaniac: Do we want log messages? I found them extremely helpful during development, and with multiple folder structures I think they would be beneficial during deployment as well.

bin/build.sh
```diff
@@ -2,17 +2,46 @@
...
   pandoc | # Markdown -> HTML
   wkhtmltopdf --encoding 'utf-8' --user-style-sheet 'assets/style.css' -T 15mm -B 15mm -R 15mm -L 15mm --quiet - "$pdf_name" # HTML -> PDF
 done
 
-# Move pdf files to build directory
+compile_date=$(date '+%Y-%m-%d')
+
+# Find educational Markdown files and build PDFs out of them.
+find 'edudoc' -name '*.md' | while read file; do
+  echo "Processing $file, edudoc style"
```
suushiemaniac: For log output, please see my other comment at L7: https://github.com/thewca/wca-documents/pull/120/files#r344507366

bin/build.sh
viroulep: I definitely don't mind having that, the travis log is pretty short anyway.

bin/build.sh
```diff
@@ -2,17 +2,46 @@
...
+  pandoc -s --from markdown --to html5 --metadata pagetitle="$document_title" "$file" | # Markdown -> HTML
+  wkhtmltopdf --encoding 'utf-8' --user-style-sheet 'assets/edudoc-style.css' -T 15mm -B 15mm -R 15mm -L 15mm --header-html "$header_html" --footer-center "[page]" --quiet - "$pdf_name" # HTML -> PDF
+done
+
+# Remove potentially cached PDFs from last build run
 rm -rf build
-cp -r documents build
-find build -name '*.md' -exec rm {} +
-find documents -name '*.pdf' -exec rm {} +
+# Create build dir so that the following cp operations maintain folder structure
```
viroulep: For anyone else reading this: this requires some slight changes on the WCA website side of things, on which I'll work.

suushiemaniac:
I have added a fix for embedding images, because the path resolution used by ẁkhtmltopdf` is not that intuitive. c/p from my code comment:

```bash
```

This means we now produce an intermediate `html` file during build, which is then removed later on during that same build process.

cubewhiz:
Let us know when this is ready to be merged so that the Board can provide the approving review.

viroulep:
I just pushed @leonopulos updates, it's not finished though.

viroulep:
Superseded by #154.

PR #142, opened 2020-01-30
WCRP version 1.1

cubewhiz:
Candidate Delegates have been renamed to Junior Delegates. The type of Delegate role in these cases is not necessary in temporary cases as Candidate (now Junior) Delegates and Full Delegates have the same rights and duties at competitions.


cubewhiz:
Can be merged once the Board vote is passed.

PR #143, opened 2020-01-30
WCA Probationary Status Policy v1.2

cubewhiz:
Candidate Delegates have been renamed to Junior Delegates. In this case, the type of Delegate is irrelevant, and our definition of Delegate includes Candidate (now Junior) Delegates anyway.


cubewhiz:
Can be merged once the Board vote is passed.

PR #148, opened 2020-02-02
Competitor tutorial edits

Samuel-Baird:
Event:
> Change "there are 18 official events." to "there are 17 official events."

Inspection:
> Change "lifts the hands from the timer" to "lifts their hands from the timer"

Solving: 
> Change "Choosing when to start the solve it’s up to the competitor and they can do that at any point during the inspection phase." to "Choosing when to start the solve is up to the competitor and they can do that at any point during the inspection phase."

Stopping the solve:
> Change "The judge checks if the puzzle is solved or not, and confirms the correctness to the competitor" to "The judge checks if the puzzle is solved or not, and tells the competitor the result"

Important regulations, stopping the timer:
The misalignment image shouldn't be here, maybe telling the competitors cannot touch the puzzle after the solve can replace this?

Formats:
> Change "The most common ones are:" to "The legal formats are:"

Important definitions:
> Replace "Combined round" with "Cutoff round"
> Change "If you managed to get at least one resolution below the cutoff, those solves already count as the beginning of your average/mean, and you can continue with the round." to "If you managed to get at least one result below the cutoff you can continue with the round and finish your average/mean."
> Change "DNS: Did Not Start. DNS means that the competitor did not start the attempt." to "DNS: Did Not Start. DNS means that the competitor was eligible for the attempt but declined it"
> "This means that you get the chance of redoing an attempt using an extra scramble because there was an incident during the original attempt"


leonopulos:
This PR can be closed, once [PR120](https://github.com/thewca/wca-documents/pull/120/) is merged. It is just a relic of the time when I didn't have the permissions to commit directly to the other PR.
All the changes that were done here are also in the other PR now.

viroulep:
Superseded by #154.

PR #149, opened 2020-02-26
Update Email Account Policy

cubewhiz:
Clarify that email signatures should not contain logos or unnecessary text.


cubewhiz:
Wait for Board vote to conclude before merging.

PR #156, opened 2020-04-27
Update Delegate Probationary Status.md

### LINKED ISSUE #151: Delegate Probation and Gear Team Orders
When Gear team orders are made the WCA takes a financial risk before the Delegate pays for the equipment. This is generally within risk appetite - the risk is taken with WCA Delegates who are generally trustworthy. However, I don't think the WCA should be taking this risk with Delegates who have previously made late payments on WCA Dues and are still on probation from that.

I think this could be currently put as a restriction when the WFC places delegates on probation without any changes to policies, but an update to the Probation policy and the Dues policy could be helpful. 

Additionally, there is not currently a way to place Delegates on probation if they fail to pay Gear invoices on time. I think giving the WFC the ability to do this in the probation policy would be helpful should it ever become an issue.

Epride:
Fixes #151

### DIFF
--- a/documents/policies/internal/Delegate Probationary Status.md
+++ b/documents/policies/internal/Delegate Probationary Status.md
@@ -1,6 +1,6 @@
 # WCA Delegate Probationary Status Policy
 <div class="version">
-### Version 1.2
+### Version 1.3
 </div>
 
 ## Policy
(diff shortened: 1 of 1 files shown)


Epride:
Await Board vote before merging

PR #168, opened 2020-08-11
Wcrp 3.0

HSavich:
Submitting for final review

### DIFF
--- a/documents/policies/external/Competition Requirements.md
+++ b/documents/policies/external/Competition Requirements.md
@@ -31,7 +31,6 @@ An official WCA Competition must:
       4. Advising other competition officials when necessary.
       5. Deciding about disqualifying competitors during the competition.
       6. Generating, providing, and maintaining the security of the scramble sequences ([WCA Regulation 4b2)](wca{regulations/#4b2})).
-      7. Making a copy of the WCA Regulations available at the competition.
    3. The WCA Delegate may delegate some of their responsibilities to organizers or other officials.
 9. Comply with the rules of “qualified sponsorship” set forth by Publication 598 of the Internal Revenue Service (IRS), as well as any additional rules or laws in the local country of the competition.
    1. The competition website may announce and display the business name or logo in connection with the competition.
@@ -42,7 +41,7 @@ An official WCA Competition must:
 10. Competitions with proceeds going to a charity must indicate so on the competition website, preferably in the Information section of the 'General info' tab. The name of the charity must be included. Any such charity needs to be recognized in the country where the competition is held.
    1. Having this indicated on the competition website must be done as soon as the charity of the proceeds for has been decided. This must be done prior to the competition taking place.
 
-If WCA Regulations are not correctly observed during a competition, results may be modified at the discretion of the WCA Delegate, WCA Regulations Committee, WCA Results Team, WCA Disciplinary Committee, or the WCA Board.
+If WCA Regulations are not correctly observed during a competition, results may be modified at the discretion of the WCA Delegate, WCA Regulations Committee, WCA Results Team, WCA Disciplinary Committee, WCA Ethics Committee, or the WCA Board.
 
 ### Announcement Criteria
 The WCA Delegate must submit the following information when requesting approval of a competition:
@@ -69,7 +68,7 @@ The WCA Delegate must submit the following information when requesting approval 
    12. Names should not contain any unnecessary parts in the name, like "competition".
    13. The competition nickname and ID must resemble the competition name.
        1. The competition nickname and ID should only be modified if the competition name is too long (more than 32 characters). Exception: groups of competitions may have modified nicknames and IDs to ensure consistency with past and future competitions.
-    14. Abbreviations and acronyms should be avoided unless the competition name is too long (more than 32 characters).
+   14. Abbreviations and acronyms should be avoided unless the competition name is too long (more than 32 characters).
 2. Location(s)
    1. A location consists of the country, city, and venue.
    2. The location should also include subregions of the country that are common to include with addresses (e.g. the state a U.S. city is located in).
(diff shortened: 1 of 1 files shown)


documents/policies/external/Competition Requirements.md
```diff
@@ -153,6 +156,7 @@ The WCA Delegate must submit the following information when requesting approval
       1. A competition must be located in the region where the championship type is selected.
       2. A competition may have multiple championship types.
       3. A championship type for a region must only be assigned to one competition in a year.
+         1. Exception: Two competitions may have the same Championship status if one of the competitions has only 3x3x3 Fewest Moves as an event and the other competition does not hold the event 3x3x3 Fewest Moves. The two competitions must be held within two months of each other. This only applies to National Championships.
```
Baiqiang, line 159: Why 3x3x3 Fewest Moves is the special one instead of others?

Baiqiang: Can a 3x3x3 Fewest Moves simultaneous competition be a National Championship?

Epride: Yes 3x3 simultaneous fewest moves competitions will be allowed as a national championship for that event. As this change is to allow for simultaneous fewest moves competitions to be recognised as championships that is why it is restricted to just fewest moves at the moment

Jambrose777: Does this include competitions like FMC World (worldwide), FMC Euro (continental), and FMC Americas (multi-continental) to count as national championships for some countries? Or does it specifically need to be ie FMC Poland and restricted to 1 country in locations?

a-Maass: I don't think this is a good idea to do ATM, I have seen people disagreeing with it. For example Christopher Chi said "I don't see why FMC is the odd one out. If time/scheduling is an issue, why not do this for MBLD as well then?", and Glib Vedmid has said "I disagree totally, because as someone who is only competitive in FMC (national champion level I mean), and as someone who only attends Nationals in Ukraine as I live in another country, I would hate to only attend an FMC only event, as I wouldn’t afford attending both comps"

Also, what's to stop someone from just holding only one FMC venue in remote Alaska, to try to have a easier chance of podium?

c-goodyear: @Jambrose777 I think it reads as not having multi-country comps as part of this. Additionally, this would then allow competitions such as FMC Europe and FMC (World) to become the relevant championship competitions. If those competitions could crown national champions too it would be rather strange (as an aside, would the DB be able to store which countries used a multi-country comp as their national champs?). 

@a-Maass FMC would be the odd one out because it is the only event where multiple venues can handle the same scrambles simultaneously. This allows greater community accessibility to countries with more than 1 Delegate.

The only people being 'disadvantaged' by such a proposal would be very rare cases such as the one you pointed out of people who travel to their country once a year for the national champs. I think this is offset by the benefit competitors at the championship competition would get by having 3 hours in the schedule used in a different way.

For the Alaska question, do you think that the Delegates of the U.S. would accept, or promote this? In fact, I'd be extremely surprised if WCAT wouldn't question such a competitions submission.

Baiqiang: @c-goodyear `This only applies to National Championships.` So FMC Europe can't be European Championship.

I have a smiliar question that can [Cross-strait FMC](https://www.worldcubeassociation.org/competitions?utf8=%E2%9C%93&region=all&search=cross-strait+fmc&year=all+years&state=past&from_date=&to_date=&delegate=&display=list) be a Greater China Championship (both include China, Hong Kong, Macau and Taiwan)?

c-goodyear: Sorry, I should clarify. I mean FMC Europe could become this in future. If it's already allowed to be a National Championship as well this could become very confusing.

EdHollingdale: @c-goodyear A single competition can be multiple championship types in the database (e.g. WC2019, WC2015, and WC2013), but that doesn't mean WCAT would necessarily approve having e.g. FMC Europe act as a national championship for multiple European countries (and I don't think it should personally).

@Baiqiang I'd say it would be up to WCAT to determine if a Greater China Championship was covered under this regulation - but I think it should be covered in this case.

For context, we are currently intending to use this clause for [Australian Nationals 2021](https://www.worldcubeassociation.org/competitions/AustralianNationals2021) (see FAQ for that comp), with the main competition taking place 1 week after a multi-location FMC competition. I think it works well for large countries with multiple delegates and multiple major population centers (e.g. Australia). It means the National championship can more easily fit into 3 days (reducing costs) and reduces some venue challenges (don't need a large side-room), while still having a very fair means of crowning a champion in FMC. 

A lot of the "crazy" theories about how this could be used miss that WCAT still have to approve competitions, and approve all championship types, and WCAT have at least a small amount of sense.

sebastianotronto: I like the idea, but I don't see why it should be for FMC only.
It would be nice, for example, to have FMC and MBLD (and possibly 4BLD and 5BLD) in a separate competition, since they are slower and less spectator-friendly events. This takes away the hurdles of organizing them in a larger competition.
I dislike the possibility of the championship title being given in a multi-location competition, but this is another matter.

I would propose the following change:

Exception: Two competitions may have the same Championship status if **they have no event in common**. The two competitions must be held within two months of each other. This only applies to National Championships.

Ivan-Brigidano: > Exception: Two competitions may have the same Championship status if **they have no event in common**. The two competitions must be held within two months of each other. This only applies to National Championships.

@sebastianotronto but this would allow to (for example) hold 2x2 to 7x7 in one competition and the rest events in a different competition. As fas as I understand it, the flexibility that Fewest Moves has (i.e. multiple locations) is the low logistics that it requires compared to any other event, including Big Blind.

Epride (approved):
Thanks WCAT for your work on this

PR #183, opened 2020-08-17
Motion Amendments September 2020

Epride:
Set of changes to the WCA Motions for Staff review. A summary of changes will be sent to Staff shortly


documents/motions/07.2020.1 - Registered Speedcubers.md
```diff
@@ -24,5 +24,5 @@ Persons who are Registered Speedcubers have the role of Non-voting Member of the
    1. Registered Speedcubers must behave according to the Spirit of the WCA within the WCA Community.
    2. Registered Speedcubers must comply with the Regulations before, during, and after a WCA Competition that they register for and/or participate in.
 7. Revocation of registration as a Registered Speedcuber
-   1. It is not possible to remove the registration of a Registered Speedcuber in the WCA.
-   2. Acknowledged WCA Competition results and Registered Speedcuber profiles should not be removed.
+   1. The registration of a Registered Speedcuber may be resigned from the WCA by notification in writing to the Secretary.
```
Jambrose777, line 27: What does resignation from the WCA as a registered speedcuber mean? Is this the same as the anonymization process outlined in the next point (if so, should this be contact WDPC instead of the Secretary)?

Epride: This is a change of the Motions to match the Bylaws (Specifically 2.4 resignation and termination) As we don't specifically use memberships or membership categories at the moment this doesn't really change how anything is done, just makes our documents match

documents/motions/16.2020.1 - Disputes.md
```diff
@@ -1,20 +1,20 @@
...
 1. All acts and decisions of WCA Staff may be subject to disputes by affected individuals.
    1. Disputes regarding the WCA Regulations must be directed to and handled by WCA Regulations Committee, and are bound by the provisions of the WCA Regulations.
       1. Appeals regarding the final decision must be directed to and handled by WCA Board of Directors.
-   2. Disputes regarding the WCA Bylaws and the Motions must be directed to and handled by WCA Board of Directors, and are bound by the provisions of the WCA Bylaws and the Motions and the applicable governing laws.
-      1. Appeals regarding the final decision must be directed to and handled by WCA Ethics Committee.
-   3. All other Disputes must be directed to and handled by WCA Disciplinary Committee, and are bound by the WCA Regulations, the WCA Bylaws and the Motions, and the applicable governing laws.
-      1. Appeals regarding the final decision must be directed to and handled by WCA Ethics Committee.
+   2. Disputes regarding the WCA Bylaws and the Motions should be directed to and handled by WCA Board of Directors, and are bound by the provisions of the WCA Bylaws and the Motions and the applicable governing laws.
```
mirandapedro, line 14: Any particular reason why this was changed? Are there situations in which these disputes shouldn't be handled by the party described in each section?

Epride: Yes there are some cases where a conflict may exist for example if someone were to bring a dispute about how the Board handled the motions/Bylaws then it would be impractical for the Board to handle this case. Likewise generally if a member of the committee is involved in the dispute then it shouldn't be handled by that committee and it usually isn't. This change is mainly just to make the motions match what actually happens in practice

mirandapedro: Thanks!

bruchemrfe (commented):
I do not like that this is going to be removed:
   2. Acknowledged WCA Competition results and Registered Speedcuber profiles should not be removed.

The next point does not cover this. I think we should not remove results. It is like changing the past and has bad effects (for example no winner of a competition),

bruchemrfe (commented):
I understand the change but I am not sure this is right, because it has the effect that board is influencing directly who has voting rights.

bruchemrfe (commented):
I am not sure this is correct. It feels like being forced to dig your own grave. For staff it is different because they are enployees and may choose to leave the organisation.

bruchemrfe (commented):
Isn't the Bylaws a separate document? I think there is a Bylaws document and there are Motions.

bruchemrfe:
> @bruchemrfe I can't tell which parts some of your comments are directed toward. Can you either quote the relevant changes or add your comments directly within the files so that we can properly address them?
That is weird. I thought I actually did comment to the specific files. But I will check it again.

documents/motions/07.2020.1 - Registered Speedcubers.md
bruchemrfe: I do not like that this is going to be removed:
2. Acknowledged WCA Competition results and Registered Speedcuber profiles should not be removed.

The next point does not cover this. I think we should not remove results. It is like changing the past and has bad effects (for example no winner of a competition),

bruchemrfe (commented):
I understand the change but I am not sure this is right, because it has the effect that Board is influencing directly who has voting rights.

documents/motions/09.2020.1 - Senior Delegates.md
```diff
@@ -4,21 +4,21 @@
...
 # Motion
 
 Senior Delegates are WCA Staff and Full Delegates, with an additional role to manage the WCA Delegates in an attributed area.
 
 1. Senior Delegates have all the rights and duties of Full Delegates and extra rights and duties as described in this Motion.
 2. Rights of Senior Delegates
-   1. Senior Delegates can appoint Registered Speedcubers as Trainee Delegates or Junior Delegates in their attributed area.
-   2. Senior Delegates can promote Trainee Delegates to Junior Delegates in their attributed area.
-   3. Senior Delegates can promote Junior Delegates to Full Delegates in their attributed area.
-   4. Senior Delegates can appoint Delegates as Regional Delegates in their attributed area.
```
bruchemrfe, line 18: I understand the change but I am not sure this is right, because it has the effect that board is influencing directly who has voting rights.

documents/motions/10.2020.7 - Ethics Committee.md
```diff
@@ -28,5 +28,5 @@ The WCA Ethics Committee is an Advisory Committee of the WCA.
       2. Requests for appeal by the WCA Board on behalf of a Registered Speedcuber or WCA Staff.
       3. Financial Support of the WCA Ethics Committee by the WCA Financial Committee.
       4. Questions and reports to the confidant within the WCA Ethics Committee.
-2. All WCA Staff must fully cooperate with investigations by the WCA Ethics Committee.
+2. All WCA Staff and Registered Speedcubers must fully cooperate with investigations by the WCA Ethics Committee.
```
bruchemrfe, line 31: I am not sure this is correct. It feels like being forced to dig your own grave. For staff it is different because they are enployees and may choose to leave the organisation.

documents/motions/09.2020.1 - Senior Delegates.md
cubewhiz: This is just a clarification of the existing process. From Motion 08.2020.1:
6.3 Senior Delegates decide at any time which persons are proposed as WCA Delegates, then
send in an application for these persons to the WCA Board of Directors.
6.3.1 The WCA Board of Directors rejects or approves the application.
6.3.2 After approval, the Senior Delegate appoints the persons as Junior Delegates.

9.2.2 To promote a Junior Delegate to Full Delegate, the Senior Delegate must submit the
corresponding application to the WCA Board.
9.2.3 The WCA Board rejects or approves the application.
9.2.4 After approval, the Senior Delegate appoints the person as Full Delegate.

There is a similar description in other documents for promotion of Trainee and Regional Delegates, all requiring Board approval.

documents/motions/10.2020.7 - Ethics Committee.md
c-goodyear: Since WEC only deal with Staff incidents, this shouldn't be the case that Registered Speedcubers are "digging their own grave" because WEC wouldn't be investigating them. One thing to note, is that if WEC becomes non-functioning, then WDC would take over their role. If the WDC became non-functioning it would make sense to have WEC perform their role, therefore this clause would make sense.

@Epride maybe the WDC / WEC motions should be updated to show who is taking on WDC responsibilities in case of WDC being non-functioning. I would assume it's WEC, but I can see the case being made for Board as well.

cubewhiz: This serves two purposes:
1. Usually WEC investigates Staff conduct, so in most cases a Registered Speedcuber would be a witness, not the person being investigated.
2. In cases in which the WEC is investigating a WDC matter (due to a conflict of interest, for example), this is already a requirement for WDC investigations:

> 3. All Registered Speedcubers must fully cooperate with investigations by the WCA Disciplinary
> Committee.

This just eliminates the discrepancy of having the process handled in two different ways depending on who conducts the investigation.

documents/motions/09.2020.1 - Senior Delegates.md
cubewhiz: > Isn't the Bylaws a separate document? I think there is a Bylaws document and there are Motions.
> https://www.worldcubeassociation.org/documents/WCA%20Bylaws.pdf
> 
> So it may be legally so that Bylaws are different from Motions.

The Bylaws themselves already describe how they should be amended:

> Except as otherwise provided by law, these Bylaws may be amended or repealed and new bylaws
adopted by approval of the Board, or by a supermajority (⅔) vote of the membership at a properly
announced meeting.

Currently, the Bylaws say they can be amended one way, but the Motions say the Bylaws are amended by a different process. We are keeping the amendment process from the Bylaws only so that the Bylaws and Motions do not contradict each other.

documents/motions/08.2020.2 - Trainee Delegates.md
```diff
@@ -25,24 +25,24 @@ Trainee Delegates are Community Members who have been appointed by the WCA to tr
...
+   2. Trainee Delegates shall sustain knowledge and skills regarding the Mission, Spirit, Policies and Regulations of the WCA.
    3. Trainee Delegates shall accept and comply with the decisions of the WCA Board, WCA Officers, and WCA Teams/Committees.
    4. For every WCA Competition that they are overseeing, Trainee Delegates:
       1. represent the WCA
       2. make sure Competition Organizers and Competition Staff are fully capable, equipped, and prepared to make the competition successful
       3. make sure the competition is run according to the Mission, Spirit, and Regulations of the WCA
-      4. timely submit a report about the competition to the community of WCA Delegates
-      5. timely send in the results for the competition to the WCA Results Team
+      4. contribute to the timely submission of a report about the competition to the community of WCA Delegates
+      5. contribute to the timely submission of   the results for the competition to the WCA Results Team
```
EdHollingdale, line 35:
```suggestion
      5. contribute to the timely submission of the results for the competition to the WCA Results Team
```
Remove extra spaces

c-goodyear (approved):
Vote passed.

PR #185, opened 2020-08-29
Update to CoE 5.4

c-goodyear:
Specifies written permission is required from the Board of Directors to have the WCA named in a contract.


cubewhiz:
Be sure to also update the version number of the CoE with this change.

Epride (approved):

PR #191, opened 2020-09-10
Organizer guidelines

leonopulos:
This PR is adding the markdown files for organizer guidelines. The generated pdfs still need to be made available on the website in a similar PR to [#4816](https://github.com/thewca/worldcubeassociation.org/pull/4816).

The user experience should be a page with [this text](https://docs.google.com/document/d/1z4_SYHejf41AOerSAj5VmhLum0Siojuzep33Sp6G_8I/edit?usp=sharing) that links to the individual pdfs. It should be available through a new third "Organizer-Guidelines" folder icon on the [education page](https://www.worldcubeassociation.org/education).

WQAC's initial idea was to get rid of the current implementation of [this page](https://www.worldcubeassociation.org/organizer-guidelines) (because it is barely linked and hard to find on the website) and replace it with the landing page for our new content. However, I now realized that the current /organizer-guidelines page is translated into many languages, and because of that I think it is worthwhile to keep around for a quick and easy overview of organizing.

Probably the best course of action would be to:
- extend /organizer-guidelines with our [landing page text](https://docs.google.com/document/d/1z4_SYHejf41AOerSAj5VmhLum0Siojuzep33Sp6G_8I/edit?usp=sharing) at the top, while keeping the old organizer-guidelines below that
- add a link to /organizer-guidelines on the /education page.

If @thewca/wqac members have any other input, or should I remember anything wrong from previous discussions, please let me know.

I'm documenting this future course of action mostly for myself, but I would highly appreciate help on this task from @thewca/software-team members. :)


jonatanklosko:
Hey, I believe the "Running systems" document is missing here.

Side note: any specific reason for breaking this into separate files instead of one document with a table of contents? On one hand I guess it's easier to maintain separate files, but on the other with a single document it's easier to add/remove/rename sections without updating references to specific files. I'm not saying to prefer either approach, just curious if that's been considered =)

DanielEgdal:
Hi Jonatan. Nicely spotted. I have now added the running systems document (along with a few other fixes) in https://github.com/thewca/wca-documents/pull/204. (Due to some changes in the team I created a new PR. So I'll close this PR in favour of the other).

The reason we chose to manage the set of organizer guidelines in separate files was for maintenance reasons of the Individual documents. We also did so as there are some documents we believe were more important than others so it should be easier for the reader to navigate between the documents when we have separated them (also to avoid the reader of being overwhelmed by one long document).

PR #205, opened 2020-11-24
Allow Restriction-less Competitions

Epride:
Allow competitions to be run without following the recommendations within the WCA Competition Safety Policy

Awaiting Board vote before merging


c-goodyear (approved):
Vote passed.

Issue #197, opened 2020-10-13
Bylaws Formatting


saranshgrover:
The newest Bylaws pdf has formatting issues in `2.10 Annual Meetings`. Seems like it is happening due to it being in 2 separate pages. Formatting is fine in the markdown. @jonatanklosko would you know what might be causing this?

![image](https://user-images.githubusercontent.com/24918614/95802338-34489b00-0ccb-11eb-8321-464d3345da4f.png)

denthebro:
I tried fixing the issue through the markdown file itself in this pull request: https://github.com/thewca/wca-documents/pull/207
Just added a page break

Nevseros:
@thewca/software-team Can you please close this one? It was resolved.

notrichardpeng:
sure :)

PR #207, opened 2020-11-27
Small edits to the Competition safety policy

denthebro:
Added 4 periods and fixed 1 sentence

### DIFF
--- a/documents/WCA Bylaws.md
+++ b/documents/WCA Bylaws.md
@@ -92,6 +92,8 @@ An annual meeting of the members shall take place in the month of October, the s
   4. Approval of the minutes of the previous general meeting;
   
   5. Activity reports;
+
+  <div class="page-break"></div>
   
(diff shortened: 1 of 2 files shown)


denthebro:
Second commit is related to this issue: https://github.com/thewca/wca-documents/issues/197

Epride:
For the competition safety policy the Board discusses changes every two weeks and we draft our thoughts into one document and then create a single PR to make the voting for amendment easier. As such could you leave your suggestions as issues, this is easier for us to handle.

Also the way Bylaws and Policies are handled are different so they defiantly should not be in the same PR

denthebro:
@Epride should I redo this PR as an issue and a PR for the bylaws change then?

c-goodyear:
This is being handled in #232 now and avoids the edits you made to the By-laws causing issues.

PR #237, opened 2021-06-05
Migrate to GitHub Actions

gregorbg:


denthebro:
The documents won't need to be updated in any way after this migration, right? It will just continue to produce the pdf files just like Travis used to?

gregorbg:
Yes, correct. It is simply a drop-in replacement for Travis.

gregorbg:
Note to WST: You can verify under `Settings > Secrets` that the private key `DEPLOY_PRIVATE_KEY` is installed as repository secret.

gregorbg:
Merging now, as this is blocking a whole lot of other stuff (and chances are I will have to adjust this anyways)

PR #241, opened 2021-06-13
Scramble accountability policy & update to the styling and the builder

denthebro:
This PR adds the scramble accountability policy. There is an issue I ran into however: tables are not currently supported in wca-documents. For this I had to update the builder and the stylesheets. There is only one issue left to be fixed right now, so I'll need someone else's help with this. Below is an explanation of all of the changes (besides the policy itself).

1. Added max image width to style.css and edudoc-style.css to have them sized properly
2. Added table styling to style.css and edudoc-style.css
3. Styling relevant to the edudoc header is added directly to the html, which is necessary due to the builder changes. It's only one line that's updated, so I think this is okay.
4. Added .warning class to edudoc-style.css, which the WQAC will use for styling our edudoc markdown files and a .bordered-image class to style.css, which is used in the scramble accountability policy markdown
5. Updated the body of style.css to make it consistent with edudoc-style.css (Note: I think it would be nice if we could use SASS in wca-documents and have it compiled into .css files, as there are several styles that are reused between the two css files, which means they would have to be updated together, which is harder to keep track of. I believe that I could implement this myself if given the green light).
6. (lines 13, 22, 23, 39, 49, 50) Updated the builder to include styling in the process of converting from .md to .html instead of from .html to .pdf. wkhtmltopdf being unable to include table styling properly seems to be a [known issue](https://github.com/wkhtmltopdf/wkhtmltopdf/issues/2469).
6.5. (lines 15, 16, 23) The builder script for the documents section is also changed to be consistent with the one for the edudoc section, except the header and footer are not included, as they are not needed.

The build of the scramble accountability policy ends up as expected, and other documents don't seem to be affected too much.

This solution has one remaining issue though, which I believe should have a simple solution, I just don't know how to program in shell. The issue is the path to the stylesheet. This works just fine in the edudoc section, as all files there are on the same depth: edudoc/project_name/file.md, so we can just specify the path as "../../assets/edudoc-style.css". However, the files in the documents section are on different depths, so the path I specified in there ("../../../assets/style.css") works fine for the scramble accountability policy and other files that are 3 levels deep, but not for files 1 or 2 levels deep. I have tried getting the absolute path all the way from root, but that doesn't seem to have worked for me.

Please advise how I could get the right path to style.css in the documents section. Also, please let me know if there are any issues with my solution to the styling of tables.


denthebro:
@thewca/software-team Please advise

DanielEgdal:
@thewca/software-team (the tag above didn't seem to work)

assets/edudoc-header.html
```diff
@@ -3,11 +3,11 @@
...
 </head>
-<body>
-<div style="border-bottom:1px solid black; text-align:center; width:100%; height:100%">
-    <div style="float:left; text-align:left">DOCUMENT_TITLE</div>
-    <div style="display:inline-block; margin:0 auto"><img src="../../assets/WCAlogo_notext.svg" height="50" alt="WCA logo" /></div>
-    <div style="float:right; text-align:right">Version: DATE</div>
-</div>
-<div style="padding-bottom:10px"> </div>
+<body style="margin: 0; padding: 0; font-family: arial; font-size: 18px;">
+	<div style="border-bottom:1px solid black; text-align:center; width:100%; height:100%">
```
gregorbg, line 7: Nit: The indentattion is not consistent with the `head` tag above.

bin/build.sh
```diff
@@ -10,13 +10,21 @@ find 'documents' -name '*.md' | while read file; do
...
   # deliberately NOT using the pipe | operator here because wkhtml tries to resolve image links
   # based on the location of its input file. If piping is used, wkhtml assumes /tmp as location
   # and links like "./dummy-image.png" become "/tmp/dummy-image.png" instead of "edudoc/$project/dummy-image.png"
   sed -r "s#wca\{([^}]*)\}#$wca_url\1#g" "$file" | # Replace wca{...} with absolute WCA URL.
-  pandoc -o "$html_name" # Markdown -> HTML
-  wkhtmltopdf --encoding 'utf-8' --user-style-sheet 'assets/style.css' -T 15mm -B 15mm -R 15mm -L 15mm --quiet "$html_name" "$pdf_name" # HTML -> PDF
+  pandoc -s --from markdown --to html5 --css="$stylesheet"  --metadata pagetitle="$document_title" "$file" -o "$html_name" # Markdown -> HTML
+  wkhtmltopdf --encoding 'utf-8' -T 15mm -B 15mm -R 15mm -L 15mm --quiet "$html_name" "$pdf_name" # HTML -> PDF
+  
+  #ORIGINAL
```
gregorbg, line 25: Don't check this in to the repository code please. We can always dig up earlier states of the command through git history.

bin/build.sh
```diff
@@ -27,17 +35,19 @@ find 'edudoc' -name '*.md' | while read file; do
...
+  
   file_headline=$(head -n 1 "$file")
   document_title=$(echo "$file_headline" | sed -E "s/#+\s*//")
 
   sed -E "s#DOCUMENT_TITLE#$document_title#g" "assets/edudoc-header.html" |
   sed -E "s#DATE#$compile_date#g" > "$header_html"
 
-  pandoc -s --from markdown --to html5 --metadata pagetitle="$document_title" "$file" -o "$html_name"  # Markdown -> HTML
-  wkhtmltopdf --encoding 'utf-8' --user-style-sheet 'assets/edudoc-style.css' -T 15mm -B 15mm -R 15mm -L 15mm --header-html "$header_html" --footer-center "[page]" --quiet "$html_name" "$pdf_name" # HTML -> PDF
+  #pandoc -s --from markdown --to html5 --metadata pagetitle="$document_title" "$file" -o "$html_name"  # Markdown -> HTML
```
gregorbg, line 47: see above regarding removing these commented-out lines

bin/build.sh
```diff
@@ -10,13 +10,21 @@ find 'documents' -name '*.md' | while read file; do
 
   html_name="${file%.md}.html"
   pdf_name="${file%.md}.pdf"
+  stylesheet="../../../assets/style.css"
```
gregorbg, line 13: resolve to an absolute path by using `realpath` command please

bin/build.sh
```diff
@@ -27,17 +35,19 @@ find 'edudoc' -name '*.md' | while read file; do
 
   pdf_name="${file%.md}.pdf"
   html_name="${file%.md}.html"
-
   header_html="${file%.md}-header.html"
-
+  edudoc_stylesheet="../../assets/edudoc-style.css"
```
gregorbg, line 39: see above regarding realpath. Also, why the heck does this have two `..` but above it has three?!

gregorbg (changes requested):
Thanks for handling all the styling updates and adjustments! Some minor remarks that should be easy to fix.

In particular I'm **very** nitpicky about the `../../../assets/style.css` references. Relative paths that move up in the directory tree (the `../` part) are always **very** dangerous. Use `realpath` to navigate absolute trees **and** assume that the script is being executed in the git repository root.
If you look at the previous `wkhtmltopdf` invocation, you will notice that it uses `assets/style.css` instead of `../../assets` because the CI executes the script from within the base directory.

gregorbg:
Also, one more nitpick: Split the styling and CSS updates into a separate PR if possible,

bin/build.sh
denthebro: That's my bad, I left it in by accident

bin/build.sh
denthebro: Because in one case it's two levels deep and in the other it's three

denthebro:
> 
> 
> Also, one more nitpick: Split the styling and CSS updates into a separate PR if possible.

Are you sure? They need to be merged at the same time anyways

denthebro:
@gregorbg I have reverted the stylesheets and finally found a working solution for the path to the stylesheets. So the issue was that html doesn't understand absolute paths, it expects relative paths, but I came up with a solution where I just copy the stylesheet into each directory and have the html reference it with "./style.css" (or "./edudoc-style.css"). These stylesheets are deleted at the end just like the temporary pdf and html files. This is now a fully working solution from what I can tell, however, this means that there are various minute changes that need to be made to some of the documents, as including the styling in the html vs in the pdf results in a very slight difference, which is still noticeable at the edges of the documents. I am happy to update all of them to fit with the updated builder if you think I made a good solution. I believe that this would be better, as the wkhtmltopdf converter does not understand stylesheets as well as regular html does (hence the inability to draw tables).

denthebro:
Made some changes to some of the documents. None of this changes the contents, just tweaks them ever so slightly to fit the updated builder. This doesn't include all documents quite yet, I'll do the rest later. I also did a rebase to include the recent commits added to this repo.

1. Added line breaks in some places
2. Made line breaks in scramble accountability policy consistent with the other documents
3. Fixed numbered lists in a few documents (they don't need empty lines in between)
4. The font is ever so slightly different, mostly noticeable with the number "1" ([image](https://imgur.com/a/Tlw4ZhT) of a before (top) and after (bottom)). I don't know why this tiny difference exists, but I'm pretty sure that the newer version is the one that's consistent with the Arial font, because it's not supposed to have a line at the bottom for the number "1".
5. Unintentionally fixed tables in wca minutes with the update to the builder ([image](https://imgur.com/a/w0gBzAm)). Affected documents:
2017-12-12 WCA Board Founding Meeting.pdf
2019-08-06 WCA Board Meeting.pdf

denthebro:
Made another update to the documents. In this states ALL of the documents in the documents folder are being rendered as expected. I will try to finish up the edudoc documents next.

1. Added "the" in WCA motion 02.2021.1 - Definitions
2. Removed almost all styling in WCA motion 02.2021.1 - Definitions, as it is now included in the style.css file (had to update one line of styling to make everything consistent; [see the styling PR](https://github.com/thewca/wca-documents/pull/243)). The only styling left is the one that makes the left column not wrap around words to a new line. Ideally, we should just port this entire table to the markdown format (I can do that if requested), but for now it can stay as an HTML table, because it renders as expected.
3. Removed empty lines (they are not needed between list items and table elements)

denthebro:
Updated all educational documents to fit the updated builder (minor changes). While I was at it, I also changed page-break classes to use divs (1 line instead of 2) and to be consistent with the styling of the documents folder. [See the styling PR](https://github.com/thewca/wca-documents/pull/243).

**From what I can tell, this project is now ready for deployment**, but I would want to have at least one other person take a thorough look at it. I checked that every single document renders properly.

denthebro:
Following @gregorbg's suggestion, copied the finances folder into a new "static" folder that doesn't get touched by the pdf builder and updated the builder to simply copy all contents of this folder into the build folder at the end. They kept getting deleted after each build before this change.

PR #244, opened 2021-06-18
Major builder and styling update

denthebro:
This PR updates the builder, the stylesheets and the edudoc header, and many of the existing documents. None of the content is changed (except for a missing "the" in one spot), all of the changes are purely to do with the rendering of the PDF files. I have checked **ALL** documents to make sure the rendering is done correctly.

The point of this update is to improve the styling of the documents. Namely, it now renders tables correctly (which it didn't do before, see [comparison image](https://imgur.com/a/w0gBzAm)). The only exception was one document that had a table manually written with HTML, but I think we could update that one to regular markdown too after this is merged. I believe that there were two WCA minutes that had tables rendered incorrectly, as well as the competitor tutorial. These are rendered as expected now. All of the changes are explained here:

1. edudoc-header.html - included the only relevant styling directly in the body tag, as this file won't have access to the stylesheet, and I don't believe it really needs to
2. style.css - specified font family, set a max image width to ensure that images are rendered correctly, added table styling and added a class for images with borders, which I will use in another project soon
3. edudoc-style.css - added the same table styling as style.css and also made the page-break classes consistent with the page-break class in style.css
4. build.sh - most important update of this PR. This makes the rendering consistent between the documents section and the edudoc section, the main difference being a reference to a different stylesheet and the inclusion of a header in the edudoc renderer. The converter also no longer includes the CSS in the conversion step from HTML to PDF, but instead in the converstion step from MD to HTML. This makes the styling render properly now. This is a [known issue](https://github.com/wkhtmltopdf/wkhtmltopdf/issues/2469) with the wkhtmltopdf converter, but pandoc doesn't have the same problem.
5. Fixed an issue with the finances folder, which now resides in a new "static" folder that from now on will accommodate all PDF files that do not require any conversion. The builder simply copies this folder into the build folder at the end.
6. Page breaks have been updated everywhere in order to make them all consistent with each other. In some places page breaks were added where there wasn't one before or one moved to a more appropriate position. Page breaks can now easily be done in both the documents section and the edudoc section by putting this line at the point where a page break is needed: <div class="page-break"></div>
7. Empty lines were removed in all **lists**. List items are not supposed to have empty lines in between.
8. motions/02.2021.1 - Definitions.md - Updated styles, because only one of the styles is needed now: the one that makes sure text in the left column doesn't wrap around; the rest of the styling is included in the style.css file now. Ideally we should convert this to .md at some point.
9. running-systems.md - Changed the header tags to make them more consistent with other organizer guidelines and make the document's layout a little nicer


denthebro:
Updated the builder to copy all contents of the "static" folder into the "build/documents" directory to accomodate https://github.com/thewca/worldcubeassociation.org/pull/6259.

denthebro:
Just did a second review of every single document and it appears to be converting correctly everywhere

gregorbg (changes requested):
Pouring `static` into `documents` by default does not seem like a future-proof approach if we wanted to include static edudoc files later (or even more document groups on top of `documents and `edudoc`)

Two possible suggestions:
1. Copy everything from `static` into a target folder which is also called `static` and adapt Saransh's PR
2. Move `static` to be a subfolder of `documents` and adjust the final cleaning step in the buildscript to something like `find documents/ -name "*.pdf" -path static -prune -o -delete` (this is an impromptu **untested** proof-of-concept command based on the `man` page of `find`. Test before copy/pasting into your PR!)

bin/build.sh
```diff
@@ -1,43 +1,59 @@
 #!/bin/bash
 
+# Makes it so that all commands are displayed as they get executed
 set -x
 
 wca_url="https://worldcubeassociation.org/"
+stylesheet=$(realpath "./assets/style.css")
```
viroulep, line 7: This assumes that we're running the script from the root folder of that repository; this seems fine because we're actually assuming something similar below.
However if at some point you want to refer a file relative to the script's location, you can consider using `$(dirname "$0")` to get the script's folder and build the path from that.

bin/build.sh
```diff
@@ -1,43 +1,59 @@
...
   echo "Processing $file, vanilla style"
 
   html_name="${file%.md}.html"
   pdf_name="${file%.md}.pdf"
+  
+  file_headline=$(head -n 1 "$file")
+  document_title=$(echo "$file_headline" | sed -E "s/#+\s*//")
+  
+  # Copy stylesheet to be used by the temporary HTML files
+  cp "$stylesheet" "$(realpath $(dirname "$file"))"
```
viroulep, line 20: I don't think you need to copy the file every time at all!
Can't you just pass the `$stylesheet` path directly to pandoc?

bin/build.sh
```diff
@@ -1,43 +1,59 @@
...
 
   # deliberately NOT using the pipe | operator here because wkhtml tries to resolve image links
   # based on the location of its input file. If piping is used, wkhtml assumes /tmp as location
   # and links like "./dummy-image.png" become "/tmp/dummy-image.png" instead of "edudoc/$project/dummy-image.png"
-  sed -r "s#wca\{([^}]*)\}#$wca_url\1#g" "$file" | # Replace wca{...} with absolute WCA URL.
-  pandoc -o "$html_name" # Markdown -> HTML
-  wkhtmltopdf --encoding 'utf-8' --user-style-sheet 'assets/style.css' -T 15mm -B 15mm -R 15mm -L 15mm --quiet "$html_name" "$pdf_name" # HTML -> PDF
```
viroulep, line 19: What was wrong with this; it seemed to be working well so what is the motivation  for the changes?

bin/build.sh
```diff
@@ -47,11 +63,16 @@ mkdir build
 # Copy generated content into build folder
 cp -r documents build/
 cp -r edudoc build/
+# Copy all contents of the static folder into the build/documents folder
+cp -r static/*/ build/documents/
```
viroulep, line 67: Good news: this is unnecessary :)
See my comment [here](https://github.com/thewca/worldcubeassociation.org/pull/6259#issuecomment-864375467).

viroulep (changes requested):
Thanks for contributing here too @denthebro!

I definitely can't comment on the restyling as I assume it comes from discussion with the relevant team/committees, but the changes to the build script look very weird to me.
If I'm not mistaken the only needed change is actually just setting the document's title in the metadata for the generated pdf in `documents`.

However I would suggest a slight improvement to the build script to keep the sources untouched, now that we also have pdf files in the source tree:
  - first of all, start by creating a brand new `build` folder, containing a clean copy of the source tree (basically, move [these lines](https://github.com/thewca/wca-documents/blob/89efe7658c5d6b71f16342a513f5c1499f83193e/bin/build.sh#L43-L49) at the beginning of the script).
  - proceed with the regular build process (it may actually be possible to DRY up this part by using an empty header file for the sources in `documents`, and having a single `process_file` function or something like that). If I'm not mistaken we would just have to change the path of the `find` command to `build/documents` instead of `documents`.
  - cleanup the build folder from any non-pdf file (basically keeping [these lines](https://github.com/thewca/wca-documents/blob/89efe7658c5d6b71f16342a513f5c1499f83193e/bin/build.sh#L50-L52) untouched).
  - remove [these lines](https://github.com/thewca/wca-documents/blob/89efe7658c5d6b71f16342a513f5c1499f83193e/bin/build.sh#L53-L57) as they won't be necessary anymore.

Hopefully what I wrote here makes sense to you, don't hesitate to reach me if anything seems unclear!

viroulep:
Ah sorry I didn't see Gregor's suggestion when posting my review!
I think doing the build process within the `build` folder is a worthy change, it also means that there would be no need for any `static` folder or something like that, and that pdf could live in their relevant folders.

gregorbg:
@viroulep Why exactly did you leave WST? :D

denthebro:
Thank you guys for all of the suggestions, but I'll be busy until the start of next week unfortunately. I'll get back to this and make all of the relevant changes as soon as I can.

bin/build.sh
denthebro: Nope, I tried this for ages, but HTML just doesn't understand absolute file paths, so you have to give it a relative path next to itself, so I just copy the file everywhere. Either that or I'm not smart enough about how to force HTML to see an absolute path from root.

bin/build.sh
denthebro: As I mentioned in my PR, wkhtmltopdf doesn't apply styling properly, as mentioned [here](https://github.com/wkhtmltopdf/wkhtmltopdf/issues/2469) as well. This makes the tables not render as expected. Pandoc doesn't have the same issue, so I applied styling in there and it worked flawlessly and even seems to have fixed a couple other minor quirks.

denthebro:
I have taken all of your suggestions into account and made a new commit. Unfortunately, it's not as elegant as I was hoping for, as the existing builder relied on "documents" and "edudoc" being in the root directory. In order to avoid fixing a bunch of docs yet again, I decided to just make temporary copies of them and then copy them into the build folder once the conversion is complete. Still, it seems like a more elegant solution than what we had before.

bin/build.sh
viroulep: Thanks for the clarification!

gregorbg (changes requested):
I don't want this code to be merged using the current `tempdocuments` and `tempedudoc` solution. That seems much too hacky and I am sure there must be a better way.

What goes against copying both the `documents` and `edudoc` folder directly (i.e. without renaming) into the `build` folder and then adjusting the compilation script so that everything happens directly within the build folder in the first place?

denthebro:
> I don't want this code to be merged using the current `tempdocuments` and `tempedudoc` solution. That seems much too hacky and I am sure there must be a better way.
> 
> What goes against copying both the `documents` and `edudoc` folder directly (i.e. without renaming) into the `build` folder and then adjusting the compilation script so that everything happens directly within the build folder in the first place?

There is most certainly a better way, but it involves the documents again, because they have hard-coded references to the wca logo file in them by going one or two levels up and then into the assets folder (example from WCA Bylaws: "../assets/WCAlogo_notext.svg"). I totally don't mind going back and updating all of these references. I think it should be fine, given that I've already made minor adjustments to many documents so that they fit this updated builder.

denthebro:
@gregorbg I have updated the builder again. I made it so that you don't need any temporary folders in the root directory, and I circumvented the logo issue by just temporarily copying it into "build" at the beginning (it gets deleted at the end anyways).
If you think this looks good, I'll update the README again and make another check of every document.

gregorbg:
Thanks! There's no need to check the PDF documents manually. Thanks to a genius idea by Philippe, you can now look at the build output from the GitHub CI run by downloading the generated files in the "Artifacts" dropdown (see screenshot)
2021-06-23-092640_2480x458_scrot

bin/build.sh
```diff
@@ -1,57 +1,71 @@
...
+
+# Remove potentially cached PDFs from the last build run and make an empty build folder
+rm -rf build
+mkdir build
+# Copy source folders into build
+cp -r documents/ build/
+cp -r edudoc/ build/
+# Copy WCA logo into build/assets temporarily
+mkdir build/assets
+cp assets/WCAlogo_notext.svg build/assets/
```
gregorbg, line 17: What about jsut copying the entire `assets` folder into `build` and not only the logo?
That way you could also reference the CSS stylesheets from within the build folder rather than copying them around later manually.

gregorbg (changes requested):
Aaaalmost there, great job! Just hold on a little bit more for some final changes :D

bin/build.sh
denthebro: @gregorbg this would not work, as every single HTML file references the stylesheet and because HTML doesn't understand absolute paths, it's much better to just copy the stylesheet into each folder temporarily and have the HTML reference it directly. In fact, I haven't found any way around this. Keep in mind that this is exactly what we did with the edudoc header to begin with, and all of these files get deleted in the end anyways, so I think it should be okay.

denthebro:
> Thanks! There's no need to check the PDF documents manually. Thanks to a genius idea by Philippe, you can now look at the build output from the GitHub CI run by downloading the generated files in the "Artifacts" dropdown (see screenshot)
> 2021-06-23-092640_2480x458_scrot

That's indeed very nice, but I still have to go through all of these to make sure they rendered properly. Let me know if everything looks okay, and I'll update the README and check all of the docs.

gregorbg:
As you can see from my recent review comment, there is just one concern about the `assets` folder pending.

bin/build.sh
gregorbg: I just did a real basic test that reveals HTML _can_ handle absolute paths.

denthebro:
Updated the builder a lot. Now we are using a function that can be reused for new potential document folders in the future. All relative paths have been removed, we're only using absolute paths now. WCA bylaws and the competitor tutorial received a few very minor tweaks. Edudoc header has been updated to remove the relative reference to the logo. A {logo} shortcut has been added, which the builder replaces with an absolute path to the logo.

edudoc/competitor-tutorial/tutorial.md
```diff
@@ -3,7 +3,8 @@
 Scroll to the bottom of the document for translations.
 
 ## WCA Organization - General Introduction
-![WCA logo](../../assets/WCAlogo_notext.svg){.logo height=133px}
+![]({logo}){.logo height=133px}
```
gregorbg, line 6: Tooltip `WCA logo` is gone...?

edudoc/competitor-tutorial/tutorial.md
```diff
@@ -21,7 +22,9 @@ If you are not sure about something, always ask the Delegate!
 - **Runner**: Runners are responsible for moving puzzles between scramblers and judges.
 - **Scrambler**: Scramblers apply official scramble sequences on competitors' puzzles. They certify with their signature that the applied scramble was correct.
 
-## Objects at the competition {.page-break-before}
+<div class="page-break"></div>
```
gregorbg, line 25: Why are you inserting the `<div>` manually now instead of the curly braces tag that was used before?

edudoc/competitor-tutorial/tutorial_se.md
```diff
@@ -1,13 +1,12 @@
 # Deltagarinstruktioner för WCA-tävlingar
 
 ## WCA som organisation - Generell Introduktion 
-![WCA logo](../../assets/WCAlogo_notext.svg){.logo height=133px}
+![]({logo}){.logo height=133px}
```
gregorbg, line 4: `WCA Logo` tooltip, see above

documents/minutes/2019-08-06 WCA Board Meeting.md
```diff
@@ -21,8 +21,8 @@ The Board agreed to schedule quarterly Board meetings. Attendance at such meetin
 
 As a result of negative feedback regarding the timing for applications with the current round of Continental Championships, in conjunction with feedback from Ethan Pride regarding the difficulty of planning the World Championship in less than one year, the Board agreed to open application periods earlier and extend their length to offer organizational teams more time to apply and give the successful applicants more time to prepare. The following timetable was agreed upon:
 
-| **Type** | **Year** | **Location** | **Applications open** | **Applications close** | **Board decision** |
-| :--------- | :--------- | :--------- | :--------- | :--------- | :--------- |
+| Type | Year | Location&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; | Applications open | Applications close | Board decision |
```
gregorbg, line 24: Whoopsie? What is this?

documents/WCA Bylaws.md
```diff
@@ -6,7 +6,7 @@
 <div style="margin: 200px auto; font-size: 1.8em;">
 # World Cube Association
 </div>
-<img src="../assets/WCAlogo_notext.svg" style="width: 60%; margin: auto;">
+<img src="{logo}" style="width: 60%; margin: auto;">
```
gregorbg, line 9: You can change this to a standard Markdown image reference `My image` and add the styling properties via curly braces

gregorbg (changes requested):
The buildscript is top notch now, thank you so much for your efforts!

I don't understand some of the detail changes you did within the MD documents.
Notably, you replaced the curly-braces style element properties by manual `div`s in a lot of places. I have commented on one but the same applies to almost every other file. Why have you changed this?

edudoc/competitor-tutorial/tutorial.md
denthebro: @gregorbg wait, that's a tooltip? On my render it just showed up as text. Why is it different for me?

denthebro: @gregorbg I don't think it is, I just checked on the documents that's on the website right now

edudoc/competitor-tutorial/tutorial.md
denthebro: I thought it would make a lot more sense to unify the page break styling of the two stylesheets. It doesn't make sense to have a separate page-break-before and a page-break-after. The page-break class is just page-break-before now, and the page-break-after class is gone. Also, I think it's a lot more intuitive for people to insert a line that visually shows where the page break is, instead of thinking that page-break means it'll break the page above. I think this is cleaner, although we could add another shortcut like {pb} that would be changed to the proper page-break div using sed during the build process. What do you think?

edudoc/competitor-tutorial/tutorial_se.md
denthebro: Replied above

edudoc/competitor-tutorial/tutorial.md
gregorbg: A tooltip is the small box that pops up next to your mouse cursor if you hover over an element. Probably depends on the PDF viewer whether you can see it.

In any case, I'd prefer if you leave the information that is currently in the documents and don't modify it, even if it might seem redundant in your viewer.
We may choose to remove redundant tooltips in another PR, it is simply not part of the purpose of this commit.

documents/minutes/2019-08-06 WCA Board Meeting.md
denthebro: Yeah, this is the result of a very long and tedious struggle to get the proper sizing of all table columns here. I tried just messing around with the number of dashes for a while here, but it doesn't seem to work well, because the table is too wide or something. This hack ended up making the table the most compact out of everything I tried. If you can find a way to keep it as compact as this without adding a bunch of spaces, then please do, haha. I'm not sure how to size it properly tbh.

edudoc/competitor-tutorial/tutorial.md
gregorbg: No sed magic where you don't need it. sed makes sense for file paths because they are complicated and can also change at runtime. CSS classes cannot.

I am mostly opposed to manually inserting div tags when pandoc has a feature to do div styling for you. Having page-break-before and page-break-after gives you more fine-grained control about the layouts of your document; I believe this is a good thing. Same argument as above: The purpose of this PR is to update and unify the build logic. If you want to discuss page breaks, open a separate PR. But please keep them in curly braces inside this commit.

documents/WCA Bylaws.md
denthebro: I just kept it as it was done originally, but sure, I can change this to markdown

edudoc/competitor-tutorial/tutorial.md
denthebro: Sure, I'll put it back, I just assumed that because my builder is rendering it as text, it would work the same in production

documents/minutes/2019-08-06 WCA Board Meeting.md
gregorbg: How does the table look if you don't add all the non-breaking spaces?

documents/WCA Bylaws.md
gregorbg: Fair point, but in this case it is obvious that the img tag is exactly equivalent to a Markdown image  reference

edudoc/competitor-tutorial/tutorial.md
denthebro: @gregorbg okay, that's fair. Should I update it in edudoc-style then? Because the way it was done originally, it wasn't consistent between the two documents.

documents/minutes/2019-08-06 WCA Board Meeting.md
denthebro: The column is narrower and there's a line (or two) where text wraps because of that, which makes the table take up more space, when it doesn't have to.

edudoc/competitor-tutorial/tutorial.md
gregorbg: I have no strong preference as long as you keep them in curly braces or pandoc ::::: div blocks.

denthebro: @gregorbg I will revert it to how it was originally then and update the edudoc stylesheet. Just fyi, I tried using the pandoc div blocks, but they must always take up at minimum two lines, which is why I replaced the page breaks with these ones everywhere. Looking back at it though, I do agree that it's still easier to type in the ::::: version.

documents/minutes/2019-08-06 WCA Board Meeting.md
gregorbg: I'd argue that a table not being perfectly aligned is less ugly than manually squeezing in non-breakable spaces in the source code.

If you really, _really_, *really* wanted. there is a thing called "CSS" to solve styling problems for you ;-) But this should not be your concern now, just leave it without the nbsp and if the Board feels they want prettier tables in their not-so-super-relevant meeting minutes they can approach WST and we will figure something out

edudoc/competitor-tutorial/tutorial.md
denthebro: @gregorbg wait, I just realized that I got that the other way around. It was edudoc that had the two separate classes and style.css had one (page-break-before). What should I do then? Should I make edudoc-style.css consistent with style.css and update all uses of the classes to use ::::: instead of my manual html divs?

gregorbg: I believe you should fulfill the following two criteria:
1. Offer both page-break-before and page-break-after for fine-grained user layout control.
2. Only insert manual `<div>` blocks where necessary. Use the pandoc Markdown extensions (i.e. ::::: and curly braces) wherever possible.

denthebro: Done

denthebro:
Updated the page-break styling and made minor adjustments to documents

gregorbg (approved):
LGTM!

denthebro:
I just looked at the output of the builder from Github and it really does seem like the tooltips show up as simple text (see WCA logo in the beginning of the competitor tutorial). On the website right now the tooltip doesn't show up either when pointing on the image. I think that it's probably best to leave it empty after all.

denthebro:
I

denthebro:
I have updated the README, added header tags for the version label in a few internal policy documents to make them consistent with all other documents, added missing period in 2019-10-22 WCA Annual Meeting and fixed a minor styling issue in the competitor tutorial. I have now compared every document with the original docs on the website and the only remaining issue is that wca logo tooltip that shows up as text in the competitor tutorial and its translation. I think the simplest fix is to just leave it empty.

denthebro:
I believe this PR is now fully finished. As mentioned previously, all of the new document renders are consistent with the ones on the website. The only visible differences are the tables, which are now rendered correctly, an added missing period in one sentence and page breaks in a few places. I only changed page breaks in places where there was a header at the very bottom of the page, so I bumped it to the next page or in places where the now fully rendered tables took up more space, so the page breaks had to be adjusted accordingly.

gregorbg (approved):
LGTM, finally! :D

gregorbg (approved):

c-goodyear (approved):
Everything looks good from the Board's perspective.

PR #272, opened 2022-01-23
Add Russian translation of competition templates

dmint789:
I translated the document and included a link to it in the English document


KerrieJarman (approved):
Approving as per Deni's request

PR #275, opened 2022-03-02
WCRP update to 4.2

c-goodyear:
Clarify what to do when a venue does not have or list a maximum capacity.

PR #276, opened 2022-03-02
Specify a competition can only be part of one series.

c-goodyear:


Epride (approved):

c-goodyear:
This undid the wording addition of #275 since it's the same file. @gregorbg do you have a recommendation on the easiest fix for this?

gregorbg:
As far as I can tell, both this change and the change from #275 are now part of the `master` branch.
When browsing through the WCRP source code (as of writing this comment) I can both see the remark about being part of only one Series as well as the remark "If such a maximum number is not available, the best estimate from a Delegate may be used."

Go to https://github.com/thewca/wca-documents/blob/master/documents/policies/external/Competition%20Requirements.md and CTRL+F search for "A competition may only be part of" and "If such a maximum". Both points should be there.

c-goodyear:
Ah yes, it's also appearing in the site version now. No issue then! :-)

PR #304, opened 2022-05-01
Update Vision and Strategy

Epride:
Update Vision and Strategy document on the WCA to the new effective version. I believe this should do the job but would appreciate someone checking my work


c-goodyear (approved):
Looks like this will fix it to me.

PR #310, opened 2022-08-28
Add Armenian translation of the judge tutorial

dmint789:
This PR adds the Armenian translation of the judge tutorial made by Mayis Jraghatspanyan


dmint789:
I think that was the final force push, the output file looks right. Please approve.

gregorbg:
Do we have any review of whether his translation is correct? There are no verified translators for `am` on the website currently...

dmint789:
We do not. I read the translation back into English using Google Translate and it seems to check out, although this method obviously does not convey certain nuances of language.

DanielEgdal:
Given Deni has checked it through google translate, and Mayis has been a listed organiser for 100% of the competitions in his country I think this is enough verification that the translation is suitable for the website.

gregorbg (approved):
I consider Daniel's post above as an approval.
Let's see if our build toolchain can handle Armenian Unicode! :smile:

PR #315, opened 2022-10-19
Re-word and add info about not using the WCA's non-profit status to receive lower rent prices

AWSMcube:
Got confused the first time I read it, had a chat with the WEC to clear up questions. Re-writing so it doesn't happen again


nsilvestri (changes requested):
I am fine with these changes. I've seen a lot of confusion about the WCA's nonprofit status and how that does (not) relate to competitions, so a clarifying change like this is useful.

As an old PR there are some conflicts that must be resolved but once that is done I can approve.

PR #316, opened 2022-10-28
WCRP Updates November 2022

c-goodyear:
See commit headers for individual details.

### DIFF
--- a/documents/policies/external/Competition Requirements.md
+++ b/documents/policies/external/Competition Requirements.md
@@ -123,7 +123,7 @@ The WCA Delegate must submit the following information when requesting approval 
             1. Missing the registration deadline.
             2. Not having enough money to pay the registration fee.
    2. The registration requirements must include the registration period(s) with a clearly defined time range with a start and end time and date.
-      1. Each registration period must be announced publicly on the WCA website at least 48 hours before it starts.
+      1. The first registration period must be announced publicly on the WCA website at least 48 hours before it starts.
          1. An active registration period may be extended after request of the Delegate.
          2. The competition's WCA Delegate(s) and/or members of the organization team may register for the competition prior to the first registration period, however their registration(s) must not be approved until the first registration period has started.
       2. Different registration periods may have different registration requirements (e.g. restricting which events a competitor can register for).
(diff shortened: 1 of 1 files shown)


documents/policies/external/Competition Requirements.md
```diff
@@ -73,25 +73,26 @@ The WCA Delegate must submit the following information when requesting approval
...
          1. Time frames when both are held cannot overlap.
    5. Series competitions
-      1. Series competitions are a special classification that can be assigned to WCA Competitions that are held less than 33 days apart for which the distance between them is less than 100 km, with the intention to accommodate more competitors to compete in one area.
+      1. Series competitions are a special classification that can be assigned to WCA Competitions that are held less than 33 days apart for which the distance between them is less than 200 km, with the intention to accommodate more competitors to compete in one area.
          1. Series competitions may be accepted if the need for competitions in an area is so great that the community will benefit from having multiple competitions held in a short period of time. 
          2. All competitions within the same series must be submitted and announced together.
+            1. A series competition may have an additional competition announced at a later time if it has identical events to another competition in the series, and if it still fulfills all other requirements for a regular competition.
```
Jambrose777, line 92: nit to match other uses.
```suggestion
            1. Series competitions may have an additional competition announced at a later time if it has identical events to another competition in the series, and if it still fulfills all other requirements for a regular competition.
```

c-goodyear: Good catch! Although upon re-reading "A series may have..." is probably more grammatically correct. Do you have thoughts on that?

Jambrose777: I agree, but the issue is probably that we define "Series Competitions" instead of "Series." and it may be better to refactor that whole part:

5. Series ~~competitions~~
1. Series ~~competitions~~ are a special classification that can be assigned to WCA Competitions that are held less than 33 days apart for which the distance between them is less than 200 km, with the intention to accommodate more competitors to compete in one area.
 1. Series ~~competitions~~ may be accepted if the need for competitions in an area is so great that the community will benefit from having multiple competitions held in a short period of time. 
2. All competitions within the same series ~~must~~ *should* be submitted and announced together.
1. A series ~~competition~~ may have an additional competition announced at a later time if it has identical events to another competition in the series, and if it still fulfills all other requirements for a regular competition.
2. *Competitions in a* series ~~competitions~~ must clearly and publicly ~~state their condition of series competitions~~ *show that they are part of a series* and *state* their special participation terms.
1. A competitor is only permitted to participate in one *competition* of ~~the~~ *a* series ~~competitions~~.
1. If a competitor competes at more than one ~~of the involved series~~ competition~~s~~ *in a series*, only the results from the first competition will be recognized, and all other results will be removed. At the discretion of the WCA Delegate, results from all competitions in the series may be removed, and they may face disciplinary action.
2. A competition may only be part of one series.

blake98x (approved):
LGTM!

documents/policies/external/Competition Requirements.md
c-goodyear: Fixed!

Epride (approved):

gregorbg:
Merging per Ethan's approval above.

PR #320, opened 2022-11-20
Portuguese (European) translations of organizer guidelines

dmint789:
This adds the Portuguese translations of **EVERY** organizer guidelines document and the images. Huge credit to António Gomes, who managed to translate these in record time. I have checked the formatting of every document and added links to them in the English versions. This also uses the new system for translations that I started in the competitor tutorial directory, as this directory would clearly become too crowded with all of the original files as well as their translations. This is a draft PR for now, as I still need to add the translated lanyard image and check that everything renders correctly.


DanielEgdal (approved):
quick scan, lgtm

PR #321, opened 2022-11-23
Fix reflection document translation section in organizer guidelines

dmint789:
I accidentally put the Translations section for this document in the Portuguese version instead of the English version


DanielEgdal (approved):
good catch

PR #323, opened 2022-12-22
Fix references/links to Motions

Nevseros:
Fixes invalid links and references to the Motions. 

As a side question, WCA Documents Motion is the only one where other Motions are just mentioned and not linked, shouldn't it be the same as in other Motions?


c-goodyear (approved):
Board approved.

PR #324, opened 2022-12-22
Add sentence about synchronize button in Delegate Crash Course

dmint789:
This PR just adds a sentence about the synchronize button to the paragraph about the WCA scrambles matcher in the Delegate Crash Course, as requested by the WRT


DanielEgdal (commented):
Shouldn't we also update "updated on" text for this section?

dmint789:
@DanielEgdal sorry for taking so long, I have changed it now

dmint789:
It should be noted that the rest of the document has been shifted down by one line, but I can't think of a way to avoid this here and it doesn't seem to be too bad anyways. See: [delegate_crash_course.pdf](https://github.com/thewca/wca-documents/files/10327174/delegate_crash_course.pdf)

DanielEgdal (approved):
A bit unfortunate yeah, but that'll have to do for now

PR #325, opened 2022-12-27
Motions Amendments - January 2023

Epride:
Amendments to the Motions as proposed by the WCA Board of Directors.


Nevseros (commented):
Line 28 in `04.2022.1 - WCA Board.md` is needed to be updated along with this change - it links to this Motion, so the link itself should be changed.

blake98x (approved):
LGTM!

PR #327, opened 2023-01-13
Changed host to our prod ip

FinnIckler:
Changing the Github action to have the prod ip as the new host. Alternatively I could also change the deploy code to use AWS Run command, like in the deploy repo.
Thoughts?

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -26,7 +26,7 @@ jobs:
 
       - name: Trigger website deployment
         env:
-          WCA_HOST: worldcubeassociation.org
+          WCA_HOST: 34.208.140.116
         run: |
           mkdir -p $HOME/.ssh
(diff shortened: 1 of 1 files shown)


danieljames-dj (approved):
LGTM. Thanks, Finn.

danieljames-dj:
> Alternatively I could also change the deploy code to use AWS Run command, like in the deploy repo.

Does it mean that we don't have to hard code the IP address? If that is the case, definitely that's better. But maybe let some experienced person comment on it, I'm bit noob.

FinnIckler:
> > Alternatively I could also change the deploy code to use AWS Run command, like in the deploy repo.
> 
> Does it mean that we don't have to hard code the IP address? If that is the case, definitely that's better. But maybe let some experienced person comment on it, I'm bit noob.

Yes, that would mean we wouldn't need to hardcode the ip address. It definitely needs to be implemented in the long run, as the run command solution will still work if there are multiple instances of the wca (but the current one wouldn't), but that is still far out.

PR #335, opened 2023-06-01
Update Travel Reimbursement

EdHollingdale:
Minor update to travel reimbursement policy proposed - the document had a mix of Treasurer only, Treasurer/Senior Member, or Treasurer/Deputy leader (which we no longer define). It is easier (and aligns with current practice) to authorise Treasurer + Senior Members for such things. 

To do in the medium term is increase documentation/governance around payment authority.


blake98x (approved):
LGTM!

PR #336, opened 2023-06-03
Update Dues System

EdHollingdale:
Update to Dues policy as discussed with Board and WFC - this should now be ready for a formal Board vote. 1) Clarifies ownership of Dues setting between Board (this policy) vs WFC, 2) Clarifies some wording/terminology, 3) reduces % of registration fees paid by Bands 1 and 2 countries to 5%. 

This will be accompanied by a refresh of which countries are in which bands.

### DIFF
--- a/documents/policies/external/Dues System.md
+++ b/documents/policies/external/Dues System.md
@@ -1,6 +1,6 @@
 # WCA Dues System Policy
 
-### Version 1.5 {.version}
+### Version 1.6 {.version}
 
 ## Purpose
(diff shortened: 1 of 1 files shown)


mirandapedro:
Hi Ed! With this new refresh, will each country/region have the opportunity to give feedback on the band they are assigned to? I know some of my colleagues in Latin America (e.g. Uruguay) would value this opportunity.

EdHollingdale:
> Hi Ed! With this new refresh, will each country/region have the opportunity to give feedback on the band they are assigned to? I know some of my colleagues in Latin America (e.g. Uruguay) would value this opportunity.

This update is to the policy, not the bands specifically. The new policy gives the WFC discretion to move countries between bands at any stage. Once the policy update is approved we will send out the proposed band recalibration. We have taken into consideration feedback already received from Delegates (particularly for Uruguay), so I don't expect it to be too controversial.

mirandapedro:
That's perfect, thank you!

blake98x (approved):
LGTM!

PR #350, opened 2023-09-08
Add business card info to crash course

nsilvestri:
This PR adds information on how Delegates can create business cards for themselves via VistaPrint, and includes links to internal assets for those business cards.


DanielEgdal (approved):
LGTM

Issue #319, opened 2022-11-11
Suspensions and other Sanctions Motion doesn't mention the CoC


Nevseros:
3.2 of the Code of Conduct explicitly mentions that its violations can result in Suspensions and other Sanctions, but Suspensions and other Sanctions Motion doesn't mention the Code of Conduct violation as the reason for possible suspensions in 2.1-2.4 with the wording "only" for other reasons.

slongbehn:
Fixed in #396. This can close.

PR #354, opened 2023-09-15
Update SaoS to reference the CoC

Samuel-Baird:
As mentioned in #319


nsilvestri:
Changes to the Motions must follow the process outlined in [13.2022.01 - Amendments of Motions](https://www.worldcubeassociation.org/documents/motions/13.2022.1%20-%20Amendments%20of%20Motions.pdf). Unfortunately this PR will have to wait until that process is completed and the amendment passes.

PR #355, opened 2023-09-15
WDC crash course

Samuel-Baird:
I'm new to using Github, so please give this a good look to make sure I formatted everything correctly!

Thanks @nsilvestri for giving me an overview of how to make this :)

PR #357, opened 2023-09-22
Creation of disciplinary crash course

Samuel-Baird:
@nsilvestri please double check :)

### DIFF
--- a/edudoc/disciplinary-crash-course/disciplinary-crash-course.md
+++ b/edudoc/disciplinary-crash-course/disciplinary-crash-course.md
@@ -1,4 +1,4 @@
-# WCA Disciplinary Crash Course
+# Guidelines for Addressing Behavioral Problems and Cheating
 
 ## Disciplinary
 
@@ -48,6 +48,7 @@ The WCA Disciplinary Committee (WDC) is entitled to apply bans and other discipl
 * Contact local law enforcement immediately if there is extreme behavior that may threaten competitor safety. Once the immediate concerns of safety have been addressed, contact the WDC to report the details of what happened.
 
 ## Online Interactions (WCA Forum/Official WCA Platforms)
+The Code of Conduct is limited in scope to WCA Competitions and online on official WCA platforms. As a result, the WDC cannot take any action in response to online incidents that happen outside of official WCA platforms.
 
 * If competitors/parents/others are being impolite or inappropriate in their online interactions related to competitions, then please do your best to respectfully handle the situation (please keep in mind that you are a representative of the WCA). You may reach out to the WDC for advice on how to best handle the situation if the problem persists or you are unsure of how to proceed.
 * If you feel a person may engage in disruptive actions or pose a threat to the safety or well-being or anyone during a WCA-sanctioned competition based on their online behavior, then please contact the WDC.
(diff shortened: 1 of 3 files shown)


nsilvestri:
Thanks for fixing some of the issues! However, since there was additional feedback on the original PR that is unaddressed here, I'm going to close this PR and redirect us back to #355.

nsilvestri:
Nevermind, the original repo was deleted so we can't continue with #355. Still, the feedback from that PR should be integrated with this PR.

edudoc/disciplinary-crash-course/disciplinary-crash-course.md
```diff
@@ -0,0 +1,98 @@
+# Guidelines for Disciplinary Incidents
```
nsilvestri, line 1: I see you renamed the document! I like this new title, and we should also update the filename and containing folder to reflect this new file name: `edudoc/guidelines-for-disciplinary-incidents/guidelines-for-disciplinary-incidents.md`

edudoc/disciplinary-crash-course/disciplinary-crash-course.md
```diff
@@ -0,0 +1,98 @@
...
+
+* Delegates should primarily address minor behavioral issues of competitors at competitions themselves using [2j](https://www.worldcubeassociation.org/regulations/#2j), [2k](https://www.worldcubeassociation.org/regulations/#2k), and [2l](https://www.worldcubeassociation.org/regulations/#2l) of the regulations and applicable sub-points.
+  * If Delegates have not first tried to handle minor behavior issues by giving warnings and disqualifying competitors under regulation 2k, then the WDC is much less likely to apply any disciplinary action. We would like Delegates to attempt to handle minor incidents at a local level first before WDC gets involved. However, the WDC would like Delegates to report noteworthy behavioral issues so the WDC can record them. The WDC is always willing to give advice and take disciplinary action as needed.
+* If any extraordinary cases occur, please describe them in the Delegate Report.
+  * Please be aware the WDC will likely contact you for further information regarding the case.
+  * If an incident is discovered after the competition, this should be added to the Delegate Report, or emailed to the Delegate Report thread.
+
+## {.page-break-before}
```
nsilvestri, line 21: Putting this page break here messes up the formatting for the bullet points, since markdown doesn't understand that the indented bullet point below is a sub-bullet alongside the bullet points above. We should either:

1. Remove this line entirely and let the text naturally flow onto the next page.
2. If you want to keep all of these points on a single page, you can add `{.page-break-before}` to the end of `## Handling Behavioral Issues of Competitors at Competitions`, although this will move the whole section to a new page and you will have a lot of white space on the first page of the doc. 

Either is fine, although I'm preferential to the first option since it's the simplest and doesn't require further fiddling with page breaks every time content is added or removed that messes up the formatting.

nsilvestri (changes requested):

edudoc/disciplinary-crash-course/disciplinary-crash-course.md
```diff
@@ -0,0 +1,98 @@
+# Guidelines for Disciplinary Incidents
+
+## Disciplinary
```
nsilvestri, line 3: This isn't really a useful section title, so I think we can remove it entirely. Then the following paragraph works just like an intro paragraph.

nsilvestri (changes requested):

nsilvestri (approved):
Approved! @dmint789 when you merge, can you Squash and Merge to compress the commits? There was a lot of interim fixes.

PR #360, opened 2023-10-19
Create Working Group Formation Policy

KerrieJarman:
Creation of policy to support formation of Working Groups

### DIFF
--- a/documents/policies/internal/Working Group Formation.md
+++ b/documents/policies/internal/Working Group Formation.md
@@ -10,7 +10,7 @@ The purpose of this policy is to provide guidelines for the formation and operat
    1. A working group is a temporary team of WCA Staff and/or community members established to address a specific project or area of interest within the organization. Working groups are tasked with generating ideas, providing recommendations, and executing initiatives related to their assigned focus area
 2. Formation
    1. Working groups may be proposed by any WCA Staff member.
-   2. The Working group must be sponsored by a team leader, a Senior Delegate or an Officer.
+   2. The Working group must be sponsored by a team leader, a Senior Delegate, a Director or an Officer.
    3.  Proposals for working groups should be submitted in writing to the Board of Directors for review and approval.
 3. Approval Process
    1. The Board of Directors will review proposed working groups and assess their alignment with the organization's mission, goals, and available resources.
--- a/documents/policies/internal/Working Group Formation.md
+++ b/documents/policies/internal/Working Group Formation.md
@@ -10,7 +10,7 @@ The purpose of this policy is to provide guidelines for the formation and operat
    1. A working group is a temporary team of WCA Staff and/or community members established to address a specific project or area of interest within the organization. Working groups are tasked with generating ideas, providing recommendations, and executing initiatives related to their assigned focus area
 2. Formation
    1. Working groups may be proposed by any WCA Staff member.
-   2. The Working group must be sponsored by a team leader, a Senior Delegate or a Board member.
+   2. The Working group must be sponsored by a team leader, a Senior Delegate or an Officer.
    3.  Proposals for working groups should be submitted in writing to the Board of Directors for review and approval.
 3. Approval Process
    1. The Board of Directors will review proposed working groups and assess their alignment with the organization's mission, goals, and available resources.
--- a/documents/policies/internal/Working Group Formation.md
+++ b/documents/policies/internal/Working Group Formation.md
@@ -28,7 +28,7 @@ The purpose of this policy is to provide guidelines for the formation and operat
    1. Members are expected to actively participate in meetings, contribute ideas and insights, and complete assigned tasks in a timely manner.
    2. Members should communicate regularly with the working group leader and fellow members to ensure progress and address any challenges.
 8. Terms of Reference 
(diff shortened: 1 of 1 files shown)


documents/policies/internal/Working Group Formation.md
```diff
@@ -0,0 +1,42 @@
...
+4. Leadership
+   1. Each working group shall have a designated leader responsible for coordinating meetings, setting agendas, and ensuring progress towards established goals.
+   2. The leader will serve as the primary point of contact between the working group and the Board of Directors.
+5. Composition
+   1. Each working group shall consist of a minimum of four (4) members, including a designated leader. 
+   2. The leader should determine the most appropriate membership for the specific working group.  The leader should consider if there should be representation from particular WCA teams or committees or if membership should be open to all WCA staff.
+   3. Priority may be given to individuals with relevant expertise or experience. 
+6. Diversity and Inclusion
+   1. The WCA is committed to fostering a diverse and inclusive environment. When forming working groups, efforts should be made to ensure representation from individuals from diverse backgrounds, with a wide range perspectives, and experiences. This includes but is not limited to factors such as race, ethnicity, gender, age, sexual orientation, abilities, and socio-economic status.
```
EdHollingdale, line 26: Should say "a wide range *of* perspectives" here

documents/policies/internal/Working Group Formation.md
```diff
@@ -0,0 +1,42 @@
+# WCA Working Group Formation Policy
+
+### Version 1.0 {.version}
+
+## Purpose
+The purpose of this policy is to provide guidelines for the formation and operation of working groups within the World Cube Association (WCA) in order to facilitate collaboration, innovation, and progress in specific focus areas.
+
+## Policy
+1. Definition of a Working Group
+   1. A working group is a temporary team of WCA Staff and/or community members established to address a specific project or area of interest within the organization. Working groups are tasked with generating ideas, providing recommendations, and executing initiatives related to their assigned focus area
```
EdHollingdale, line 10: Definition is missing *why* you should set up a working group. At the moment it just seems like another way for the Board to micromanage and restrict the work and discussions WCA Staff can do/have. 

Maybe people will be willing to go through the "hoops" of this governance if there are clearer benefits to the "costs" (the months required to get board pre-approval and the ongoing reporting. Those benefits to me are having a formal process and timelines, being able to use WCA resources in the project, getting meaningful engagement from relevant WCA teams, and having a far more likely chance of having any proposed change approved by having gone through the official process. Presumably the policy should note these benefits, otherwise it is a whole bunch of compliance where it isn't readily apparent why you'd want to do a working group "officially".

documents/policies/internal/Working Group Formation.md
```diff
@@ -0,0 +1,42 @@
...
+2. Formation
+   1. Working groups may be proposed by any WCA Staff member.
+   2. The Working group must be sponsored by a team leader, a Senior Delegate or a Board member.
```
EdHollingdale, line 13: Edge case: what if a Sponsor loses that role between proposal and agreement? or while the working group is underway. Presumably a new sponsor would need to be nominated? E.g. hypothetically the sponsor was a Senior Delegate but their term expired in the intervening 2 months it took for the Board to review their email!?

EdHollingdale: Also, I would recommend replacing "Board Member" with "Officer". We need to start laying the groundwork for non-executive directors to exist.

documents/policies/internal/Working Group Formation.md
KerrieJarman: I am not sure how much of that needs to be baked into the policy. 

The why of working groups should be done in communications. If you are not doing your work in the formalised working group you risk your work not being recognised. As an organisation we want to ensure that we are focusing on the right things at the right time. 

Standardising how working groups are formed, how they are structured and operate and how report their progress should become common practice and it should become an unquestioned truth that if you want to get something done right, you do it within a formalised structure. We need to set it up in beginning to be transparent and for there to be accountability to management (whatever that looks like) and to the whole community. 

There is a whole body of literature on this topic and I don't know that it belongs in the policy.

documents/policies/internal/Working Group Formation.md
KerrieJarman: fixed

documents/policies/internal/Working Group Formation.md
KerrieJarman: Thanks changed Board member to Officer.
I am not worried about that edge case and I don't believe it needs to be defined in the policy

KerrieJarman: also added Director

EdHollingdale: I disagree, I don't think a Director alone should be able to do it. Directors should be focussed on governance not on being the people to get things done.

documents/policies/internal/Working Group Formation.md
EdHollingdale: Fine. I'd still like to have seen something in a policy but it isn't a hill I will die on. Just as with RO's the motions have the rights (i.e. the benefits you get and why you might want to become an RO) as well as the Duties (what you need to do for that recognition). This just has the "what you need to do" and doesn't commit in policy why you'd want to do that. Of course, things like the WCRP or the Regulations are much the same :/

Epride (approved):
Board Approved

PR #362, opened 2023-10-24
Update Championship Naming Requirements

Epride:
Update Competition requirements policy so that the naming requirements of Continental and World Championships match reality


KerrieJarman (approved):
Approved by Board

PR #365, opened 2023-11-22
Add point about venue temperature to finding venue orga guideline

dmint789:
This issue was raised by some competitors at a recent UK comp, so it would be good to add this point for future organizers.


Nanush7 (approved):
LGTM!

PR #372, opened 2024-01-26
Move Deployments to S3

FinnIckler:
Due to infrastructure changes we will deploy wca-documents directly to S3 and serve them via a separate URL like https://documents.worldcubeassociation.org/documents/Code%20of%20Conduct.pdf
This will not change how they are listed and shown on the website.


FinnIckler:
I tested this locally, but we can wait to merge this until the PR in the monolith is merged

.github/workflows/deploy.yml
gregorbg, line 1: (GitHub doesn't let me comment on lines that haven't changed, so I'm commenting on the whole file instead)

Around L21, we used to use an action called "JamesIves/github-pages-deploy-action" which pushes the results of the building process onto a separate `build` branch.

The only purpose of this branch was that the `update_docs` command on the old EC2 machine can directly check out this branch when cloning, and pull the documents from there. Now that we publish the built documents to S3 instead, there's no need to also publish them to the `build` branch. That entire step can be deleted (assuming that the other monolith PR also gets merged at the same time)

gregorbg (changes requested):

.github/workflows/deploy.yml
FinnIckler: done!

gregorbg (approved):

PR #377, opened 2024-03-04
Region restrictions policy

EdHollingdale:
@thewca/board as discussed, addition of Regional Restriction policy with adjustments as suggested, for you to vote on.
Additionally, I have updated Appendix B of the Dues policy to cross reference the Regional Restrictions policy.
cc @thewca/financial-committee


KerrieJarman (approved):
Approved by Board vote

PR #382, opened 2024-03-27
Spanish Translations Improvements

Nanush7:


edudoc/judge-tutorial/judge-tutorial.md
```diff
@@ -45,7 +45,7 @@ Here are some common incidents that can result in +2 or DNF:
 
 **+2**: the competitor stops the timer without fully releasing the puzzle.
 
-**DNF**: the competitor touches the puzzle before the judge has inspected the solved state. If you are certain that no change in alignment of the puzzle was done as a result of this, **+2** can be given instead.
+**DNF**: the competitor touches the puzzle before the judge has inspected the solved state. If you are certain that no change in alignment of the puzzle was done as a result of this, **+2** can be given instead. Exception: if immediately after stopping the timer, the competitor **briefly** touches the puzzle without affecting its state, the time penalty may be lifted, at the discretion of the judge
```
dmint789, line 48: Missing period at the end. Also, is it not at the discretion of the Delegate?

dmint789 (changes requested):
Left one suggestion. Also, I have a couple more suggestions:
1. We should add a point about the judge having to check the timer is in 2 pad mode
2. We should mention that the reset button must be held down on gen 5 timers or newer (to future proof it)

edudoc/judge-tutorial/judge-tutorial.md
Nanush7: Nope, [A6c](https://www.worldcubeassociation.org/regulations/#A6c). Added period.

Nanush7:
@dmint789 Looks good?

dmint789:
@Nanush7 yeah, lgtm, I'll approve. Did you update the Spanish version accordingly? Does the current Spanish translation match the English version? Also, as a general thought, do you not think we're replicating the regulations a bit too much in this document now? It might require more maintenance that way. But if you think this is kind of necessary for a doc like this, we can leave it.

dmint789 (approved):

dmint789 (changes requested):
Wait, I just realized, the version hasn't been updated. Can you change it to 2.1 in both the English and the Spanish version?

Nanush7:
@dmint789 Done! :)

> Also, as a general thought, do you not think we're replicating the regulations a bit too much in this document now?

It's important to not give just half of the information, as it may lead to confusion. I agree we should keep it as simple as possible, but I don't think the current version replicates the regulations that much.

dmint789:
@Nanush7 LGTM. Can you remind me, do you have the rights to merge? If not, I'll squash merge this for you.

Nanush7:
I didn't know we had the right to merge, but looks like I do. :p

dmint789:
Yeah, I requested the rights last year

PR #384, opened 2024-04-03
Update WCRP to 5.0

zekemackay06:
[Update Log](https://docs.google.com/document/d/1ylego2XOrrILJrWJOyq3uz67cKi6Dp39mwmsTqb8oi8/edit?usp=sharing)

### DIFF
--- a/documents/policies/external/Competition Requirements.md
+++ b/documents/policies/external/Competition Requirements.md
@@ -1,6 +1,6 @@
 # WCA Competition Requirements Policy
 
-### Version 4.4 {.version}
+### Version 5.0 {.version}
 
 ## Purpose
 The purpose of this policy is to define the requirements and processes that a competition must meet and follow in order to be recognized as a WCA Competition.
@@ -15,25 +15,30 @@ An official WCA Competition must:
 3. Comply with all [WCA Policies](wca{documents}).
 4. Be open to all competitors and spectators who wish to attend, up to the maximum capacity stated on the website or otherwise permitted by the venue. 
    1. Any restrictions placed on competitors and/or spectators who can attend must be publicly outlined in the competition information section and approved by the WCA Board or WCA Competition Announcement Team.
+   2. Individuals suspended from attending WCA sanctioned competitions may not attend any WCA sanctioned competitions until the period of their suspension has expired (see 1.1 of Suspensions and other Sanctions).
+   3. A WCA Delegate may require individuals to leave the competition venue for behavior that is unlawful/violent/indecent/unsafe, or intentionally damages venue facilities or personal property within the venue. Delegates must take into account the seriousness of the situation and the best possible course of action. Delegates must also report any such incidents to the WDC to investigate the need for disciplinary action.
 5. Be held on a single date or on two or more consecutive days.
 6. Have at least one official event.
-7. Have at least one competitor.
+7. Have at least two competitors.
 8. Have at least one designated WCA Delegate in attendance.
    1. Exception: in case the WCA Delegate cannot attend the competition, the WCA Board may appoint a trusted person to act as a temporary WCA Delegate for the competition.
    2. The WCA Delegate for a competition is responsible for:
       1. Submitting the Delegate Report to the WCA Regulations Committee and the WCA Quality Assurance Committee regarding adherence to WCA Regulations and quality standards during the competition, the overall course of the competition, and any incidents that happened. The report must be submitted within one week of the end of the competition.
       2. Submitting the competition results to the WCA Results Team within one week of the end of the competition.
-         1. Results should be double-checked and free of errors before submission.
+         1. Results must be double-checked and should be free of errors before submission.
          2. All scramble sequences used during a competition must be sent with the results of the competition.
             1. All scramble sequences must be labeled with the events, rounds, and groups for which they were used. If there are multiple groups in a round, it is not necessary to identify which competitor was in which group.
-         3. Physical records (e.g. score sheets, written solutions from 3x3x3 Fewest Moves) must be retained for a minimum of 90 days. Digital records (e.g. scramble sequences, results) must be retained for a minimum of 7 years.
+         3. Record keeping
+            1. Physical records (e.g. score sheets, written solutions from 3x3x3 Fewest Moves) must be retained for a minimum of 90 days.
+                 1. Scanned copies may be kept instead of the physical records.
+            2. Digital records (e.g. scramble sequences, results) must be retained for a minimum of 7 years.
       3. Sending corrections of competition results to the WCA Results Team.
       4. Advising other competition officials when necessary.
-      5. Deciding about disqualifying competitors during the competition.
+      5. Deciding whether to disqualify competitors during the competition per ([WCA Regulation 2k)](wca{regulations/#2k})).
       6. Generating, providing, and maintaining the security of the scramble sequences ([WCA Regulation 4b2)](wca{regulations/#4b2})).
       7. Ensuring that all outstanding incidents after a competition are resolved.
    3. The WCA Delegate may delegate some of their responsibilities to the organization team or other officials.
-9. Comply with the rules of “qualified sponsorship” set forth by Publication 598 of the Internal Revenue Service (IRS), as well as any additional rules or laws in the local country of the competition.
+9. The information listed on the WCA Website, as well as other official WCA channels must comply with the rules of “qualified sponsorship” set forth by Publication 598 of the Internal Revenue Service (IRS), as well as any additional rules or laws in the local country of the competition.
    1. The competition website may announce and display the business name or logo in connection with the competition.
    2. The competition website may link to a website homepage or competition-specific landing page.
       1. The competition website cannot link to web pages that advertise specific products or services of the sponsor (unless it is the sponsor’s homepage).
@@ -47,7 +52,7 @@ If WCA Regulations are not correctly observed during a competition, results may 
 ### Announcement Criteria
 The WCA Delegate must submit the following information when requesting approval of a competition:
 
-1. Competition name, nickname, and ID
+1. Competition name, short name, and ID
    1. The name must not be too general. It should be related in some way to the competition itself, such as:
       1. The location of the competition
       2. The competition venue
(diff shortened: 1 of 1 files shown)


documents/policies/external/Competition Requirements.md
```diff
@@ -15,25 +15,30 @@ An official WCA Competition must:
...
       3. Sending corrections of competition results to the WCA Results Team.
       4. Advising other competition officials when necessary.
-      5. Deciding about disqualifying competitors during the competition.
+      5. Deciding whether to disqualify competitors during the competition per ([WCA Regulation 2k)](wca{regulations/#2k})).
       6. Generating, providing, and maintaining the security of the scramble sequences ([WCA Regulation 4b2)](wca{regulations/#4b2})).
       7. Ensuring that all outstanding incidents after a competition are resolved.
    3. The WCA Delegate may delegate some of their responsibilities to the organization team or other officials.
-9. Comply with the rules of “qualified sponsorship” set forth by Publication 598 of the Internal Revenue Service (IRS), as well as any additional rules or laws in the local country of the competition.
+9. The information listed on the WCA Website, as well as other official WCA channels must comply with the rules of “qualified sponsorship” set forth by Publication 598 of the Internal Revenue Service (IRS), as well as any additional rules or laws in the local country of the competition.
```
EdHollingdale, line 41: But at the competition itself the rules don't need to be followed? I'm fine with this, but I'm still not sure that there isn't a general presentation that WCA competitions are somehow organised by the WCA, where qualified sponsorship could become an issue depending on how Organisers / Delegates (noting they may be the same person) present WCA involvement.

documents/policies/external/Competition Requirements.md
```diff
@@ -56,36 +61,52 @@ The WCA Delegate must submit the following information when requesting approval
...
+         2. If any location of such competitions is closer than 10 km to any other competition, then:
             1. both competitions should be at least 5 days apart, or
             2. the other competition must not have the 3x3x3 Fewest Moves event.
-      2. A 3x3x3 Fewest Moves simultaneous competition cannot be part of any other regular competition.
-         1. Time frames when both are held cannot overlap.
-   5. Series
+      2. Every location must clearly state who is the WCA Delegate for the venue.
+            1. There must be a listed non-Trainee Delegate at each venue.
+      3. A 3x3x3 Fewest Moves simultaneous competition cannot be part of any other regular competition.
+         1. Time frames when both are held cannot overlap. There must be at least one hour scheduled between the final scheduled event of one competition and the next scheduled event of the other.
```
EdHollingdale, line 106: why do we need the time frames piece and the 1 hour restriction? To me the policy requirement was always about not being able to count the results of an FMC simultaneous competition towards two competitions (i.e. determining the podium of a local comp and the simultaneous competition). I see absolutely no issue with a location of an FMC simultaneous competition taking place in the side room of another local competition, for example.

documents/policies/external/Competition Requirements.md
```diff
@@ -94,96 +115,124 @@ The WCA Delegate must submit the following information when requesting approval
...
+      2. It must be stated what the deadline is for accepting a registrant on the waiting list.
+         1. The default deadline is the time at which registration closes.
+         2. This deadline should be the same as the deadline for refunds for competitors canceling their registration, if refunds are to be given.
+      3. The organization team may specify a limit to the waiting list. Once the limit of the waiting list has been reached, the Delegate should close the registration period. 
+   6. The registration requirements may include a policy on registration refunds if offered for the competition.
+      1. A full refund may deduct any necessary transaction fees from the registration fee.
+      2. There must be a deadline after which refunds are not issued.
+         1. This deadline should be the same as the deadline to add registered events except if the deadline to add registered events is during the competition.
+      3. Under force majeure, a full refund of their registration fee may be issued beyond the deadline at the delegate’s discretion.
```
EdHollingdale, line 182: I don't think the WCA Delegate should be the one deciding on refund conditions. Refunds are a financial decision, and competition finances should be handled by organisers. I don't think organisers should have to ask for permission from a WCA representative to apply a force majeure refund policy.

documents/policies/external/Competition Requirements.md
```diff
@@ -94,96 +115,124 @@ The WCA Delegate must submit the following information when requesting approval
...
    3. Date(s)
-      1. National Championships should not be held on the same date(s) as Continental or World championships.
-      2. Continental Championships must not be held on the same date(s) as World Championships.
+      1. National Championships should not be held on the same date(s) as Continental championships of the country’s continent.
+      2. National Championships should not be held on the same date(s) as World championships.
+      3. Continental Championships must not be held on the same date(s) as World Championships.
    4. Championships should have all the officially recognized WCA events.
    5. Championships must recognize and award champion titles and their respective prizes to the winners in the region.
-   6. Organization teams of National Championships should involve all the WCA Delegates of the respective countries, at the discretion of the Senior Delegate.
+   6. Organization teams of National Championships should involve all the WCA Delegates of the respective countries and the regional organization, at the discretion of the Senior Delegate.
```
EdHollingdale, line 231: I'd prefer to say "and/or" the regional organization here, since this is how it is applied in practice.

EdHollingdale (commented):
Good iterative changes! A few comments from me :)

documents/policies/external/Competition Requirements.md
```diff
@@ -15,25 +15,30 @@ An official WCA Competition must:
 3. Comply with all [WCA Policies](wca{documents}).
 4. Be open to all competitors and spectators who wish to attend, up to the maximum capacity stated on the website or otherwise permitted by the venue. 
    1. Any restrictions placed on competitors and/or spectators who can attend must be publicly outlined in the competition information section and approved by the WCA Board or WCA Competition Announcement Team.
+   2. Individuals suspended from attending WCA sanctioned competitions may not attend any WCA sanctioned competitions until the period of their suspension has expired (see 1.1 of Suspensions and other Sanctions).
+   3. A WCA Delegate may require individuals to leave the competition venue for behavior that is unlawful/violent/indecent/unsafe, or intentionally damages venue facilities or personal property within the venue. Delegates must take into account the seriousness of the situation and the best possible course of action. Delegates must also report any such incidents to the WDC to investigate the need for disciplinary action.
```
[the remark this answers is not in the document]
slongbehn: Hi Ed, this was added because of WDC feedback.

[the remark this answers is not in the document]
EdHollingdale: I assumed so - but I don't want to create the possibility of WDC not following the policy. I feel like the word "investigate" is a strong word - the WDC may not decide it isappropriate to open an "investigation" (e.g. because of the nature of the behaviour, or the actions already taken at the comp), or an investigation may already in progress and this is additional information to it.

Similarly, Delegates should be aware of local laws and venue restrictions. If the competition is in certain public places and depending on the local laws it may not be possible for the Delegate to unilaterally require an individual to leave. 

Additionally, now that I think about it, I'm uncomfortable with the idea that the WCA Delegate (and individual acting in their capacity as a representative of the WCA) decides on this, as this opens up the WCA to legal action for events it doesn't organise. Wouldn't it be better for the competition organiser to be the ones who decide on this?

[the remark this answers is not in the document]
slongbehn: Much of this is outside of the WDC scope, so I will not respond to those portions. 

The WDC will need to look into any of these reports and whether or not further sanctions should apply, so I believe the language is appropriate. 

I am not in favor of all organizers having the sole power to remove people from venues.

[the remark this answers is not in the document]
slongbehn: Sorry for the double post, but I am fine changing the word to “determine” over investigate if it really makes an impact.

Also, 11h already allows delegates to ask people to leave. This is just creating consistency between documents.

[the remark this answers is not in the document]
EdHollingdale: I think the wording there is up to WDC - but my suggestion with policies is don't force yourself to do something of you can envisage a circumstance where it wouldn't be appropriate.

On the language: in general the WCA has far too much emphasis on the WCA Delegate being the one to decide anything of importance. Things are structured with the idea that the organiser is a 13 year old child who is being particularly helpful at that comp but all the serious stuff is done by the Delegate - so serious tasks are therefore the domain of the WCA. But the WCA isn't here to organise competitions - we just make sure they follow a set of standards, host the results, and provide software tools to facilitate their organisation. 

The real organisers are the key Delegates, or entities like Speedcubing Australia, Cube New England, etc. They should absolutely have the ability to run their events and handle issues not related to the cubing competition aspect, while liaising with WCA representatives to ensure WCA policies are followed. It's going to take a lot of effort to update WCAs thinking and documents for this, but I at least don't want more wording to be added that makes it harder. This isn't to say that particularly helpful 13 year old cubers who are a key volunteer for a given competition don't have a role to play, but let's not lie to ourselves and say they're the organiser - most WCA Delegates are organisers. We need to do a better job of separating the two "hats" to empower whichever local cubing organisation is coordinating things, and better define the WCAs role as a global governing body. The WCA having completed consistency and in essence organising competitions is simply not scalable.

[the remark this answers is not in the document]
slongbehn: I have let Zeke know I don’t really have an opinion either way. The WDC will review each report regardless of the language and determine if the action the delegates took was sufficient or if further sanctions are needed.

I don’t agree or disagree with your other comments but documents should be consistent. The regulations already contain effectively the same language. This is a larger conversation that cannot possibly be addressed by a single point in the WCRP.

documents/policies/external/Competition Requirements.md
```diff
@@ -94,96 +115,124 @@ The WCA Delegate must submit the following information when requesting approval
...
+       1. If on-site registration was advertised and the competitor limit was reached when registration closed, on-site registrations must not be accepted.
    3. The competitor limit is per competition or per event. Per day competitor limits are not allowed.
       1. Exception: for 3x3x3 Fewest Moves simultaneous competitions, the competitor limit is per location.
    4. The competitor limit is not greater than half the maximum number of people allowed at public gatherings in the competition venue such that each competitor is able to bring 1 guest into the venue.
       1. If such a maximum number is not available, the best estimate from a Delegate may be used.
    5. The WCA Competition Announcement Team will review the competitor limit and/or qualification results or rankings to ensure that they are utilized properly.
       1. The WCA Competition Announcement Team, together with the Senior Delegate, may set a minimum competitor limit per region and event list.
+      2. Competitions with a low expected competitor turnout may be rejected at the discretion of the Senior Delegate for the region.
```
timreyn, line 141: This should be one line later.

documents/policies/external/Competition Requirements.md
```diff
@@ -94,96 +115,124 @@ The WCA Delegate must submit the following information when requesting approval
...
       2. Additional registration periods may be added at least 24 hours before they start.
+        1. A registration period may be extended and active immediately, at the discretion of WCAT. 
       3. Different registration periods may have different registration requirements (e.g. restricting which events a competitor can register for).
    3. The registration requirements must include the procedure for completing a valid registration.
-   4. The registration requirements must include the deadline for updating registered events for competitors with completed registrations.
+   4. The registration requirements must include the deadline for adding registered events for competitors with completed registrations.
       1. This deadline must not be earlier than the time registration is set to close. 
-         1. The deadline may be set to be just before an event is scheduled to take place. 
+         1. The default deadline is the same as the time at which registration closes.
+         2. The deadline for adding registered events may be left blank, in which competitors may add registered events up until an event takes place.
```
timreyn, line 168: This is a bit confusing -- line 165 says it needs to have a deadline, line 168 says it doesn't, line 170 says that in some cases it doesn't.

Would it be clearer to say that:
The registration requirements may have a deadline for adding registered events.
If there is a deadline, it must not be earlier than the time registration is set to close.
If there is no deadline, competitors may add registered events up until that event takes place.

documents/policies/external/Competition Requirements.md
```diff
@@ -94,96 +115,124 @@ The WCA Delegate must submit the following information when requesting approval
...
+      3. The organization team may specify a limit to the waiting list. Once the limit of the waiting list has been reached, the Delegate should close the registration period. 
+   6. The registration requirements may include a policy on registration refunds if offered for the competition.
+      1. A full refund may deduct any necessary transaction fees from the registration fee.
+      2. There must be a deadline after which refunds are not issued.
+         1. This deadline should be the same as the deadline to add registered events except if the deadline to add registered events is during the competition.
+      3. Under force majeure, a full refund of their registration fee may be issued beyond the deadline at the delegate’s discretion.
+      4. Competitors who cancel their registration while on the waiting list should be given a full refund of their registration fee.
+      5. Competitors who are still on the waiting list at the deadline for accepting competitors from the waiting list should be given a full refund of their registration fee.
```
nsilvestri, line 184: Out of curiosity, any reason why this isn't "must"?

documents/policies/external/Competition Requirements.md
nsilvestri: In principle I agree, but the fact that payments for most competitions are taken from within the WCA website itself (even if it's to a non-WCA Stripe account) requires the WCA to have some accountability in how refunds are processed.

If payments are taken entirely on a 3rd-party site I'd personally be fine with waiving any refund requirements.

EdHollingdale: There are already many rules in thos policy for how refunds etc need to be handled (deadlines, waiting lists, etc) which organisers need to follow. I agree that given they are using the WCA website there needs to be some quality standards applied, but I don't think this is one of them.

More generally the WCA needs to have rules that acknowledge that Organisers do all the work, and WCA Delegates ensure rules are followed (an oversight role). I think that's mostly there, but a lot of the rules are written in a way where it assumes the organiser is a 14 year old who doesn't know what they're doing and the Delegate is implicitly expected to be the one doing the work and making the key decisions (e.g. deciding to issue refunds). Maybe that works for some regions of the US, but it doesn't work when the organiser is a legal entity with significant turnover and where, quite frankly, the people making financial decisions are more competent than the average WCA Delegate. Any rules that are set which restrict organisers apply to Speedcubing Australia Inc just as much as your random organiser - so they need to be set in such a way to be flexible.

documents/policies/external/Competition Requirements.md
zekemackay06: This is a difficult point because we (WCAT) can't enforce how competitions comply with this point beyond the information on the competition website. You're absolutely right that we don't do well at showing the difference of the WCA versus organizers, but this is far out of my wheelhouse.

Epride (approved):

PR #386, opened 2024-05-12
Update the build pipeline

dmint789:
This is a big PR, but this was a very important update for us to make. It should improve the developer experience greatly when it comes to building the PDFs to see a preview of the documents, which was previously only really easy to do by creating a PR and waiting for the Github pipeline to do it, which always took a while and made it harder to iterate with minor adjustments. This PR does the following:

- Update the build script to use weasyprint instead of wkhtmltopdf, which is no longer maintained and causes bugs in the output PDFs sometimes. Some other minor improvements to this script have also been made.
- Add a new build script that uses Docker for consistent file rendering.
- Unify the duplicated styles between the documents and edudoc stylesheets, and keep the directory-specific styles in their own separate files: documents-style.css and edudoc-style.css
- Make minor adjustments to the documents themselves to ensure there are no alignment issues and to make some of the Markdown look cleaner.
- Renumber points 5. and 6. of the [WCA Bylaws](https://documents.worldcubeassociation.org/documents/WCA%20Bylaws.pdf) section 2.4 as 4. and 5. respectively (point 4. was missing). Keep in mind that it's displayed as 1-5 in production rn, because Markdown seems to ignore what number you actually put there and simply uses the correct one in order.


bin/docker_build.sh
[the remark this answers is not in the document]
dmint789: Fixed

README.md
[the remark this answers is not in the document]
dmint789: But you left a review comment. Should I just resolve it then?

dmint789:
@nsilvestri could you provide an example of an unwanted page break due to the spacing between lines? For me the spacing looks the same as it does in production rn.

dmint789:
Staff crash course fixed

nsilvestri:
@dmint789 the page breaks caused by spacing are not "unwanted", just slightly different. 1-2 lines will be pushed to the next page, or vice versa. Very minor stylistic thing, just noting it as an inconsistency.

dmint789:
Fixed the bullet point issue in the competitor tutorial

dmint789:
Okay, now I need to update the Github actions

dmint789:
Found another issue: the counters reset when there is content between two items (see the competition templates doc for an example). Currently working on fixing this.

dmint789:
Fixed

dmint789:
I have personally gone through every single markdown document in the repository, fixed all alignment issues and made sure they all look almost exactly the same as the original docs rendered with wkhtmltopdf. The PR is now ready for review.

@nsilvestri could you please review the changes in the documents and edudoc folders? It's almost exclusively alignment fixes and changes from pure HTML to markdown/pandoc syntax in some of our old docs. My expectation is that you won't find any issues, as I tried being as thorough as possible, but it's good to have another set of eyes on the changes.

@thewca/software-team could you please review the changes in the root of the repo? Namely, the changes to the assets and bin folders, the README and my newly-created Dockerfile.

I will also review all of the aforementioned changes myself for good measure.

dmint789:
@thewca/software-team also, I'll need someone to override the PR build test, cause it's failing due to not being up to date with these changes

nsilvestri (approved):
Thanks for seeing through this colossal effort. LGTM on behalf of WQAC.

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+COPY ./bin/install_dependencies.sh .
+RUN ./install_dependencies.sh
+RUN rm ./install_dependencies.sh
+
+# All of this makes it so we can build PDFs without having them owned by root
+ARG USER_ID
+ARG GROUP_ID
+RUN groupadd --non-unique --gid $GROUP_ID user
+RUN useradd --non-unique --uid $USER_ID --gid $GROUP_ID user
+USER user
```
gregorbg, line 15: Out of curiosity: The above `WORKDIR /home/app` path kinda implies that there is already a user called `app` pre-defined by the Ubuntu image. Does that one have the necessary privileges?

Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM ubuntu:latest
```
gregorbg, line 1: Have you considered using something like `alpine` instead?
Downloading a whole Ubuntu image for a couple of PDFs feels a bit overkill.

assets/edudoc-style.css
```diff
@@ -1,61 +1,80 @@
...
+    content: element(headerLogo);
+  }
+
+  @top-right {
+    content: element(headerDate);
+  }
+
+  @bottom-center {
+    content: counter(page);
+  }
```
gregorbg, line 25: Where do the variables like `headerTitle`, `page`, etc. come from?

bin/build.sh
```diff
@@ -1,73 +1,81 @@
 #!/bin/bash
 
-# Makes it so that all commands are displayed as they get executed
-set -x
```
gregorbg, line 4: Why did you decide to drop this?

bin/build.sh
```diff
@@ -1,73 +1,81 @@
...
 edudoc_header=$(realpath "./assets/edudoc-header.html")
 # Current date
 compile_date=$(date '+%Y-%m-%d')
 
-# Function for converting Markdown files from a given folder into PDF files using the given stylesheet
-# The first argument ($1) is the folder and the second argument is the stylesheet ($2)
+# Function for converting Markdown files from a given directory into PDF files.
+# The first argument ($1) refers to the directory name (e.g: documents or edudoc).
 convert_to_pdf() {
+  rm -rf "build/$1"
```
gregorbg, line 15: There is a "delete" step at the very bottom of the build script, and because of that, individually removing PDFs here wasn't necessary before. Why is it necessary now?

bin/build.sh
```diff
@@ -1,73 +1,81 @@
...
   cp -r "$1/" build/
 
-  # Find formal and legal Markdown files and build PDFs out of them.
+  # Find Markdown files and build PDFs out of them.
   find "build/$1" -name '*.md' | while read file; do
-    echo "Processing $file using the ${1} stylesheet"
+    echo "Converting ${file#build/}..."
```
gregorbg, line 20: You dropped the part that prints information about the stylesheet being used. This could make debugging harder.

bin/build.sh
```diff
@@ -1,73 +1,81 @@
...
+    # This creates the custom header for each edudoc and prepends it to each file
+    if [ "$1" = "edudoc" ]; then
       # Set the document title, path to the logo and the date
-      sed -E "s#DOCUMENT_TITLE#$document_title#g" "$edudoc_header" |
-      sed -E "s#WCA_LOGO_PATH#$logo#g" |
-      sed -E "s#DATE#$compile_date#g" > "$header_html"
+      sed -E "s#\{document_title\}#$document_title#g" "$edudoc_header" |
+      sed -E "s#\{wca_logo_path\}#$logo#g" |
+      sed -E "s#\{date\}#$compile_date#g" |
+      echo -e "$(cat /dev/stdin)\n\n$(cat $file)" > "$file"
```
gregorbg, line 41: Am I correct in reading that you're basically pre-pending `stdin` to the existing file?

This `echo` is hard to read and understand, and manually typing out `\n` line breaks is never a good idea in scripts. If this actually does what I think it does, then consider some other approach along the lines of:
1. Piping the `sed` stuff into a new. temporary file using `echo > $temp_file`
2. Appending the original file using `cat $file >> $temp_file`
3. Renaming `$temp_file` to the original `$file` if necessary.

bin/build.sh
```diff
@@ -1,73 +1,81 @@
...
     fi
 
-    # Markdown -> HTML
-    pandoc -s --from markdown --to html5 --css=$2 --metadata pagetitle="$document_title" "$file" -o "$html_name"
+    # Markdown => HTML + remove unsupported styles generated by pandoc
+    pandoc -s --from markdown --to html5 --css "$main_stylesheet" --css "$custom_stylesheet" --metadata pagetitle="$document_title" "$file" |
+    sed -E "s#gap: min\(4vw, 1\.5em\);##" |
+    sed -E "s#overflow-x: auto;##" > "$html_path"
```
gregorbg, line 47: What is this `sed` doing?

bin/build.sh
```diff
@@ -1,73 +1,81 @@
...
-    if [ $1 = "documents" ]; then
-      wkhtmltopdf --encoding 'utf-8' -T 15mm -B 15mm -R 15mm -L 15mm --quiet "$html_name" "$pdf_name"
-    elif [ $1 = "edudoc" ]; then
-      wkhtmltopdf --encoding 'utf-8' -T 15mm -B 15mm -R 15mm -L 15mm --header-html "$header_html" --footer-center "[page]" --quiet "$html_name" "$pdf_name"
+    if [ "$1" = "documents" ]; then
+      # --presentational-hints makes things like the start attribute of ol elements work,
+      # which is required for adding more content between two consecutive ordered list items
+      weasyprint --presentational-hints --encoding 'utf-8' "$html_path" "$pdf_path"
+    elif [ "$1" = "edudoc" ]; then
+      weasyprint --presentational-hints --encoding 'utf-8' "$html_path" "$pdf_path"
```
gregorbg, line 55: The two cases are completely identical as far as I can tell, so there's no need for an if-else.

bin/build.sh
```diff
@@ -1,73 +1,81 @@
...
+    # Remove the Markdown and HTML files
+    rm "$file" "$html_path"
   done
 }
 
-# Remove potentially cached PDFs from the last build run and make an empty build folder
-rm -rf build
-mkdir build
+# Create the build directory if it didn't exist in the first place
+mkdir -p build
```
gregorbg, line 64: In line with my other comment above (feel free to answer only one of them): Why did you move this `rm` step that is clearly present in the left side of the diff into the conversion method body?

bin/build.sh
```diff
@@ -1,73 +1,81 @@
...
+  convert_to_pdf documents
+  convert_to_pdf edudoc
+fi
 
-# Remove all non-PDF files and empty folders from build
+# Remove all non-PDF files and empty directories from build
 find build/ -type f -not -name "*.pdf" -delete
 find build/ -type d -empty -delete
+
+echo -e "\nBuild finished!"
```
gregorbg, line 81: If you need to echo a line break at the end of the string, something with your previous output is wrong. Never manually encode line breaks in script files, previous `echo` prints always terminate with a new line.

If some `weasyprint` commands don't terminate with a new line and you really really need this extra line break, then issue two `echo` statements with the first `echo`ing an empty string. (And probably raise an issue against WeasyPdf, terminal output commands should always be terminated with line breaks for readability.)

bin/docker_build.sh
```diff
@@ -0,0 +1,31 @@
...
+  docker rmi wca-pdf-builder
+fi
+
+if [ "$1" = "--rebuild" ] || [ "$2" = "--rebuild" ] || [ "$image_exists" = true ]; then
+  docker build --tag wca-pdf-builder --build-arg USER_ID=$(id -u) --build-arg GROUP_ID=$(id -g) .
+
+  # Make sure the command completed successfully
+  if [ $? != 0 ]; then
+    exit
+  fi
```
gregorbg, line 23: This is what the `||` control flow operator in Bash is made for `try_some_command --arg "foo" || echo 'the command failed!'`. In this context, you probably want `docker build || exit`.

gregorbg: https://linuxsimply.com/bash-scripting-tutorial/redirection-and-piping/piping/double-pipe

bin/install_dependencies.sh
```diff
@@ -1,18 +1,11 @@
...
+  echo "Installing the necessary dependencies"
 
-if ! [ -x "$(command -v wkhtmltopdf)" ]; then
-  echo "wkhtmltopdf not found. Installing..."
-  # Copied from https://github.com/thewca/worldcubeassociation.org/blob/5698831f9addb26091641c80beadfa7c902bcb18/chef/site-cookbooks/wca/recipes/regulations.rb#L17-L18
-  wget https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/0.12.4/wkhtmltox-0.12.4_linux-generic-amd64.tar.xz -O /tmp/wkhtml.tar.xz
-  sudo tar -xf /tmp/wkhtml.tar.xz --strip-components=1 -C /usr/local
+  apt update &&
+  # fonts-liberation includes Liberation Sans, which is used in this repo, and fonts-noto is the fallback for wider language support
+  apt install -y wget fonts-liberation fonts-noto pandoc weasyprint
```
gregorbg, line 8: The reason why we did not use `apt` before is that it allows us to pin the specific version of `pandoc` and `wkhtmltopdf`. That way, unexpected breaking changes can be avoided.
Your new `apt` way looks more convenient on the surface, but actually it will always just take the newest version that is available at the time of running the script. In six months or one year, that might be a new major version that suddenly needs new CLI args to be passed, or otherwise the output looks horrible. (Like, say for example the CLI parameter for defining "use this CSS file please" changes from `--css $some_css_file` to `--stylesheet $some_css_file`. In the "best" case scenario, the old `weasyprint` invocation will fail because then `--css` becomes an unrecognized CLI option. In the "worst" case, it will just continue, silently skipping the `--css` option, which means our documents get compiled without stylesheet and "suddenly" look awful and everybody will scratch their heads what happened because these issues are hard to debug.

gregorbg: TL;DR: Is there a way for you to pin specific versions of at least `pandoc` and `weasyprint`?
(The font files don't need pinned versions because we don't invoke CLI commands on them that could break in the future)

gregorbg (changes requested):
This review includes the technical scripts only. I have not reviewed every single `md` file nor have I meticulously verified the generated PDF contents. That is left entirely up to WQAC.

Dockerfile
dmint789: I have, but that uses a different package manager, so the install script cannot be used with it. This PR used debian slim before, but I figured I might as well make this image as close to the github action as possible. If you think Debian slim is close enough to ubuntu, I can change it back.

bin/build.sh
dmint789: I felt that it was unnecessary, seeing as the docker build pipeline just handles everything for you anyways. This was kinda used for debugging for you.

bin/build.sh
dmint789: I found myself sometimes waiting for a specific file to build, and this made it easier to spot it when it wasn't buried in .md and .html files, so I think this is a little better.

bin/build.sh
dmint789: You're right, and since two stylesheets can be used now, I could update this to show which stylesheets are being used.

bin/build.sh
dmint789: That's exactly how it used to work. I felt like the temporary file was unnecessary, so I changed it. I don't think this part is that bad personally.

bin/build.sh
dmint789: Yeah, so as the comment above it mentions, there is this totally unnecessary styling that pandoc generates that's not supported by weasyprint, and this sed just removes it. I have not found this affecting anything, it just removes the annoying warnings during the build process.

bin/build.sh
dmint789: Answered above

bin/build.sh
dmint789: Oops, good catch. There used to be a difference with wkhtmltopdf.

bin/build.sh
dmint789: Nono, it totally does have a line break before anyways, this just adds another line break to visually separate this line from the stuff above it.

dmint789: If you feel strongly about it, I can change this to two echo statements

bin/docker_build.sh
dmint789: Ooh, thanks for sharing! Didn't know about that one

bin/install_dependencies.sh
dmint789: Good point there. I think the best option would be to specify the package versions using apt. I'll make this change.

Dockerfile
dmint789: This is the output I get when doing ls -l /home/app:

`drwxr-xr-x 11 ubuntu ubuntu 4096 May 28 10:06 wca-documents`

assets/edudoc-style.css
dmint789: These are just custom names I chose, and they're used below inside of the `@media print` block

bin/build.sh
dmint789: Oh, but the new line characters are totally unnecessary, I'll get rid of them

Dockerfile
dmint789: But actually, I didn't find a big difference in build times between the two, and you only need to build the image once and then you can keep using it, so I'd just leave Ubuntu as the base image

gregorbg: Okay yeah then whatever. It's not about build times necessarily, more about how much space you're taking up on the host machine.

bin/build.sh
gregorbg: Hm, in that case I guess it's fine.

bin/build.sh
gregorbg: I still do feel quite strongly that `echo $(cat /dev/stdin)` is not a good idiom, especially when piping. Not sure what actual Bash gurus say about this, but unless you find a really compelling argument or an article that underlines that the way you're doing it is common best practice, I'd rather have it be separate echo statments like before.

Dockerfile
gregorbg: That does not fully answer my question. Does the user `app` actually exist or am I misreading some file/folder configuration?
If the user exists, please try out whether it has the necessary privileges for running wasyprint. I cannot figure this out by staring at `chmod`esque permission return values.

dmint789: Oh, my bad, I misunderstood. The app user doesn't exist in the ubuntu docker image, no. This directory can simply be used for your app's files.

Dockerfile
dmint789: Right. So the image with an Ubuntu base is 1.76GB. I just tried the Debian slim base image again, but it won't work, because of the newly-added package version locking for pandoc and weasyprint, since Debian and Ubuntu have slightly different versions of packages in their respective repos.

dmint789:
@gregorbg is this good to be merged now?

bin/build.sh
```diff
@@ -1,73 +1,78 @@
...
+    # This creates the custom header for each edudoc and prepends it to each file
+    if [ "$1" = "edudoc" ]; then
       # Set the document title, path to the logo and the date
-      sed -E "s#DOCUMENT_TITLE#$document_title#g" "$edudoc_header" |
-      sed -E "s#WCA_LOGO_PATH#$logo#g" |
-      sed -E "s#DATE#$compile_date#g" > "$header_html"
+      sed -E "s#\{document_title\}#$document_title#g" "$edudoc_header" |
+      sed -E "s#\{wca_logo_path\}#$logo#g" |
+      sed -E "s#\{date\}#$compile_date#g" |
+      echo -e "$(cat /dev/stdin)$(cat $file)" > "$file"
```
gregorbg, line 43: I still feel quite strongly about this `echo` with a wrapped `cat /dev/stdin` call.
What's so hurtful about a plain, simple, readable

```bash
sed -E | sed -E | sed -E > $target_file
cat $origin_file >> $target_file
```

documents-guide.md
```diff
@@ -103,7 +103,25 @@ If you are making a document that needs a version label, use this:
 ### Version 1.0 {.version}
```
gregorbg, line 103: Might be worth bumping the version?

documents-guide.md
gregorbg, line 1: For what it's worth, you're changing quite a lot of inline HTML to custom pandoc `::::`-style syntax throughout all documents.

image

It might be worth explaining the usage of `::::: { .html-class }` and similar snippets here.

documents/minutes/2017-12-12 WCA Board Founding Meeting.md
```diff
@@ -1,80 +1,82 @@
...
-</div>
+:::::
 
 On the motion and by unanimous vote, the following persons were elected temporary chairperson and secretary for the first meeting:
 
 - Temporary Chairperson: Ron van Bruchem
 - Temporary Secretary: Robert Burton
 
-<b class="text-center">MINUTE BOOK</b>
+#### **MINUTE BOOK** {.text-center}
```
gregorbg, line 23: You're changing this to become a heading (MD `####` element) whereas previously it was just "some text" wrapped in a `b` tag.

Why do you also additionally need to make it bold with `**text**`? Shouldn't headings be bold by themselves?

gregorbg: And if they are indeed not bold by default (which would be very surprising!) then have you considered making them bold as part of your custom CSS stylesheet?

bin/docker_build.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# Get the directory to build argument
+if [ -n "$1" ] && [ "$1" != "--rebuild" ]; then
+  directory_to_build=$1
+elif [ -n "$2" ] && [ "$2" != "--rebuild" ]; then
+  directory_to_build=$2
+fi
```
gregorbg, line 8: What are the two different styles of calling this script, and why are they both necessary?

bin/docker_build.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# Build the Docker image, if the --rebuild flag was used or if an image hasn't been built yet
+
+if [ -z "$(docker images | grep wca-pdf-builder)" ]; then
+  image_exists=true
+  docker rmi wca-pdf-builder
+fi
+
+if [ "$1" = "--rebuild" ] || [ "$2" = "--rebuild" ] || [ "$image_exists" = true ]; then
```
gregorbg, line 17: You can set a `$is_rebuilld` flag in the "Get the directory to build argument" block above, and then just do one "equal to true" check here

gregorbg (changes requested):

documents-guide.md
dmint789: This is not the actual version of this document, it's an example of how you can use the version class. This doc doesn't have versioning, we can just update it like the README.

documents-guide.md
dmint789: It's explained on lines 31-39

bin/build.sh
dmint789: The problem is that we need to prepend here, not append. That's what the wrapped cat call achieves here.

dmint789: Found a different, hopefully cleaner, solution using tee

documents/minutes/2017-12-12 WCA Board Founding Meeting.md
dmint789: They're bold by default, that's my bad

bin/docker_build.sh
dmint789: You can do one of the following: script, script --rebuild, script edudoc, script documents, script edudoc --rebuild, script documents --rebuild, script --rebuild edudoc, script --rebuild documents.

bin/docker_build.sh
dmint789: But that one doesn't check for the flag being == to --rebuild, it checks the opposite along with another check. I'd have to add more if statements to do what you're suggesting. Would you like me to do that?

bin/build.sh
dmint789: That actually didn't work, and it breaks every other document. Reverting it to the one line echo solution.

dmint789: Let me know if you figure out a cleaner way to prepend the header html

gregorbg: Uhm... Doesn't my original suggestion do just that?
Run some `sed`s, write them to a new, temporary file (i.e. they get added to the *beginning* of that new file) and then use `>>` to _append_ the original file to the `sed`'ed stuff.

bin/docker_build.sh
gregorbg: Okay, I didn't consider the mumbo-jumbo'd way of appending the `--rebuild` argument in odd places. Nevermind,

bin/docker_build.sh
gregorbg: Okay, in that case it should be fine as-is.

bin/build.sh
dmint789: Ohh, so you meant to append to the new file that has the header html in it. I'm dumb, I'll do that then.

bin/docker_build.sh
dmint789: I figured it would be easier for the user if the order of the arguments didn't matter

bin/build.sh
dmint789: Done

dmint789:
Looks like all issues are resolved now

gregorbg (approved):
The technical build scripts have my approval. I did not (and will not) check the final rendered output of every single document in the new pipeline.

Sidenote: The CI error log looks like you forgot `sudo` somewhere

dmint789:
Oh, maybe it's the apt commands. Also, does the build pipeline use the setup files from master or from this branch I'm trying to merge? I assumed the CI error would never go away until this PR is merged.

dmint789:
@thewca/board can you please approve? This probably requires Board approval, because I had to edit many of the documents themselves, but the only real change is the last point in the description of this PR, but it doesn't even affect the rendered file, since the numbers themselves are ignored. I just wanted to fix the numbering in the Bylaws Markdown. I have personally checked the rendering of every single document in this repo, and everything has remained consistent, bar some very minor alignment here and there.

dmint789:
@gregorbg Looks like I've answered my own question, as the PR build test has a different output after my commit now. I'll try to fix it, looks like the repos used by the github action are different from the ubuntu image repos.

dmint789:
@gregorbg according to [this](https://itsfoss.com/apt-install-specific-version/), package version locking is not possible on ubuntu. I'd just remove the pandoc and weasyprint package versions and call it a day, unless you know of a way to get a specific version of these packages from apt.

nsilvestri (dismissed):
Approved on behalf as Board regarding non-content or formatting changes to Board-sensitive documents.

dmint789:
@gregorbg the checks are finally passing. Can you please re-approve, cause the new commits made your approval stale. Maybe we can revisit package version locking in another issue and PR, cause this is currently blocking two of our projects.

gregorbg (approved):

nsilvestri (dismissed):

PR #394, opened 2024-06-30
WSAP v2 Update

dmint789:
This PR makes the following changes to the WSAP (the diff on GitHub is hard to read for some reason, but the list below mentions every change to the policy to make it easier to review):

1. ~~Give the WQAC the same responsibility as the WRC in the enforcement of this policy.~~ This has been reverted.
2.  Move the definition of terms to the top to improve the readability of the policy, remove redundant term descriptions (Competitor, WCA Delegate, Organization Team, Regulations) and add missing ones (Drop-off table, Scrambling Area)
3. Add point 1.3
4. Remove point 2.1 (it's part of the Definition of Terms now), and replace it with 2.1.1
5. Update point 2.2
6. Add point 2.2.2
7. Add new point 3.2, and make 3.1.1 a sub-point of 3.2 instead (3.2.1)
8. Remove what used to be point 3.3, which simply duplicated one of the regulations
9. Add point 3.4.2
10. Update point 4.2
11. Remove point 5.1.1, the first sentence of which is already covered by the updated point 2.2, and the second sentence of which is not reasonably enforceable by Delegates at all times
12. Add "randomly-generated" to 5.1.3 (now 5.1.2), and remove the old 5.1.3.1, which becomes redundant
13. Replace "one of the scramblers" with "a trusted scrambler" in 5.1.3.3.1 (now 5.1.2.2.1)
14. Add 4 new scrambling setup examples, making it clear what kinds of setups adhere to the policy, and what don't
15. Remove what used to be the first setup example, as it was not a very good setup to begin with, despite being presented as such
16. Resize all setup images to fit two examples per page instead of one. The document will remain 6 pages long, despite the additional setup examples.
17. Update some of the old setup example descriptions
18. Various wording tweaks


Nanush7 (approved):

documents/policies/external/Scramble Accountability.md
```diff
@@ -1,91 +1,103 @@
 # WCA Scramble Accountability Policy
 
-### Version 1.0 {.version}
+### Version 2.0 {.version}
 
 ## Purpose
 
-The purpose of this policy is to define standards of how official WCA Scrambles, which in addition to the WCA Regulations, must be treated. The policy is targeted towards the WCA Delegate and organization team. Not fully following the Scramble Accountability Policy may result in disqualification of the affected results, at the discretion of the WCA Regulations Committee.
+The purpose of this policy is to define standards on how official WCA scrambles must be treated. The policy is targeted towards WCA Delegates and organization teams. Not fully following the Scramble Accountability Policy may result in disqualification of the affected results, at the discretion of the WCA Regulations Committee or the WCA Quality Assurance Committee.
```
nsilvestri, line 7: It is preferred to have the ability to disqualify results for WSAP violations to be limited to WRC at the moment. WQAC can separately operate within its power to advise WRC in these matters but whether such a decision is in violation of Regulation 1c1 is within WRC's scope.

nsilvestri (changes requested):

documents/policies/external/Scramble Accountability.md
dmint789: But we were previously advised that this responsibility may be added for our team. Has the Board changed its stance on this topic?

dmint789: Removed mention of the WQAC

nsilvestri (approved):
Approved by Board vote

PR #411, opened 2024-08-11
Fix Inconsistency Between Regs and CoE

CarterKoala:
(Pending review of ~~WIC~~ and Board)

As requested by WRC, addresses inconsistency between the Regs and CoE:

There was a source of confusion brought up surrounding [Regulation 1c](https://www.worldcubeassociation.org/regulations/#1c), specifically this part:

> The WCA Delegate may appoint other members of the organization team to carry out specific responsibilities on their behalf, but is ultimately accountable for how these responsibilities are carried out.

Specifically, this means that WCA Delegates may appoint someone to handle incidents on their behalf. 

The Code of Ethics currently states:

> 3.1 Decisions on incidents required to be made by a WCA Delegate must only be made by those listed on the WCA competition page.
> 3.1.1 Non-listed Delegates can make decisions on incidents only with the approval of the listed Delegate(s).

This would make it seem like those that are not Delegates would not be able to handle incidents, even when appointed by the Delegate as per Regulation 1c. This potential update addresses this.

### DIFF
--- a/documents/Code of Ethics.md
+++ b/documents/Code of Ethics.md
@@ -53,7 +53,7 @@ The official communication channel for the WIC is via email. For any concerns ab
       1. WCA Staff should always be aware of their target audience (e.g. their age, experience, nationality, etc.) and their tone/language when they make statements while acting in their capacity as WCA Staff.
 3. **Conduct of Official WCA Competitions**
    1. Decisions on incidents required to be made by a WCA Delegate must only be made by those listed on the WCA competition page.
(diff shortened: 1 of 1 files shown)


nsilvestri:
Board review required before merging, bringing up internally

CarterKoala:
Thanks @nsilvestri. FYI we will email Board once I have confirmed everything on the WIC side.

nsilvestri (approved):

Issue #398, opened 2024-08-01
Document titles showing unrendered styling text in browser


nsilvestri:
Repro:

1. Visit https://documents.worldcubeassociation.org/documents/minutes/2022-10-08%20WCA%20Annual%20Meeting.pdf in browser (Firefox and Chrome have similar but slightly different issues)
2. The tab title is `::::: {.text-center} - 2022-10-08 WCA Annual Meeting.pdf` on Firefox and `::::: {.text-center}` on Chrome. Neither are correct

dmint789:
Oh gosh, it must be the recent update that affected this. I did not anticipate the tab title changing 😅

PR #414, opened 2024-08-21
Fix pagetitle for minutes and motions

Nanush7:
Fixes https://github.com/thewca/wca-documents/issues/398

I'm not a bash expert, but I think this solution should be fine.

Note that I used `(([a-zA-Z0-9.\-]|[[:blank:]])+)\.md` to match the filename. I tried `[a-zA-Z0-9.\- ]`, `[a-zA-Z0-9.\-\ ]` and `[a-zA-Z0-9.\-\s]`, but none of those worked (with docker) to match spaces in the filename. That's why I added `[[:blank:]]`.

Anyway, it can be as simple as `(.+)\.md`, but I guess the complex regex should be safer.

### DIFF
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -21,8 +21,14 @@ convert_to_pdf() {
     pdf_path="${file%.md}.pdf"
     html_path="${file%.md}.html"
 
-    # Use first header as the document title; remove all # characters and whitespace from the beginning
-    document_title=$(head -n 1 "$file" | sed -E "s/^#+\s*//")
+    if [[ "$file" =~ build/documents/(motions|minutes)/(([a-zA-Z0-9.\-]|[[:blank:]])+)\.md ]]; then
+      # We need this for minutes and motions to have the correct document title. Otherwise it uses "{.text-center}"
+      # or "SUBMISSION OF PROPOSED MOTION".
+      document_title=${BASH_REMATCH[2]}
+    else
+      # Use first header as the document title; remove all # characters and whitespace from the beginning
+      document_title=$(head -n 1 "$file" | sed -E "s/^#+\s*//")
+    fi
     # Absolute path to the custom stylesheet
     custom_stylesheet="./assets/$1-style.css"
     custom_stylesheet_path=$(realpath "$custom_stylesheet")
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -21,7 +21,7 @@ convert_to_pdf() {
     pdf_path="${file%.md}.pdf"
     html_path="${file%.md}.html"
 
-    if [[ "$file" =~ build/documents/(motions|minutes)/(([a-zA-Z0-9.\-]|[[:blank:]])+)\.md ]]; then
+    if [[ "$file" =~ build/documents/(motions|minutes)/(([a-zA-Z0-9._\-]|[[:blank:]])+)\.md ]]; then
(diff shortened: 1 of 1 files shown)


dmint789:
Hmm, I'm wondering if it would be easier to just use the intended document title as the first line, and just remove that line during the build step. OR we can straight up try using some HTML meta tag in the markdown directly, and the meta tag would include the document title.

dmint789:
Alright, I just tried the <title> tag idea, and it doesn't work. But the other idea is feasible. We could make it so it either uses the first line, if it starts with a # or it parses a line like this and removes it during the build process: `{document_title="DRAFT MINUTES OF ANNUAL MEETING OF THE MEMBERS OF WORLD CUBE ASSOCIATION"}`.

Nanush7:
I mean, if we want the _easiest_ approach, then we probably want the current proposal, which requires zero changes to the actual documents.

Any reason why you don't feel comfortable with it?

dmint789:
> I mean, if we want the _easiest_ approach, then we probably want the current proposal, which requires zero changes to the actual documents.
> 
> Any reason why you don't feel comfortable with it?

Actually, I guess your options isn't bad, but I'm not a fan of limiting it to just the motions and minutes. I think instead we should check if a normal # title exists, and if not, just use the file name, as you've done here.

Nanush7:
The problem is that a normal # title exists for motions, but it's not the title we want. `# SUBMISSION OF PROPOSED MOTION`.

dmint789:
Oh, makes sense now

bin/build.sh
```diff
@@ -21,8 +21,14 @@ convert_to_pdf() {
     pdf_path="${file%.md}.pdf"
     html_path="${file%.md}.html"
 
-    # Use first header as the document title; remove all # characters and whitespace from the beginning
-    document_title=$(head -n 1 "$file" | sed -E "s/^#+\s*//")
+    if [[ "$file" =~ build/documents/(motions|minutes)/(([a-zA-Z0-9.\-]|[[:blank:]])+)\.md ]]; then
```
dmint789, line 24: Should the [] include an underscore (_) character too?

dmint789: Also, try [[:space:]] instead of blank. This stuff is explained [here](https://fabianlee.org/2020/01/29/bash-using-bash_rematch-to-pull-capture-groups-from-a-regex/).

dmint789 (changes requested):

bin/build.sh
Nanush7: Filenames in motions and minutes don't include the "_" char, but I can add it if you think it'd be better.

I used [[:blank:]] instead of [[:space:]] because space includes many more chars that wouldn't be included in any filename anyway (afaik).

Captura de pantalla 2024-09-01 a la(s) 12 50 03

dmint789: Oh, my bad, I thought blank would be the more generic one. And yeah, I think it's best to include _ to future-proof it.

dmint789 (approved):

dmint789:
@thewca/software-team could someone from the WST please review this?

Nanush7:
@thewca/software-team ?

dmint789:
@gregorbg would you mind approving this?

gregorbg (approved):

PR #423, opened 2024-09-04
Fix format in WMCT Motion

Nanush7:
The format for [this motion](https://documents.worldcubeassociation.org/documents/motions/10.2024.14%20-%20Major%20Championship%20Team.pdf) is currently broken. As you can see, some of the fields at the top of the document are on the same line.

Also, is there any reason to use "co-ordinating" instead of "coordinating"?
1.2.3 Has "coordinating".


nsilvestri:
Thanks for fixing this. We can also use "coordinating", it's the more common variant of this word.

nsilvestri (approved):
Approved by Board vote

PR #435, opened 2024-10-16
WCRP 5.3 Update

zekemackay06:
As requested by Board, this update contains several new baseline quality standards for WCA competitions.

Changes made:
- Strengthen proximity policy to 30km, only allow two competitions within proximity distance regardless of events
- Disallow competitions in vehicles
- Heavily restrict private residences as venues
- Require a complete schedule at announcement
- Disallow competition days of more than 16 hours

Minor changes
- Fix wording of exceptions
- Require competitors to only compete at one location of an FMC simultaneous competition

### DIFF
--- a/documents/policies/external/Competition Requirements.md
+++ b/documents/policies/external/Competition Requirements.md
@@ -108,6 +108,7 @@ The WCA Delegate must submit the following information when requesting approval 
          1. Competitions held alongside 3x3x3 Fewest Moves simultaneous competitions should have a break around the simultaneous competition in case of schedule delays.
       4. A National Championship holding only 3x3x3 Fewest Moves may be a part of a simultaneous competition alongside a competition holding the National Championship status of another country.
          1. 3x3x3 Fewest Moves simultaneous competitions holding the National Championship status of two or more countries must recognize and award champion titles and their respective prizes to the winners in all the regions that take part.
+      5. A competitor must compete at the same location for all of their attempts at a 3x3x3 Fewest Moves simultaneous competition.
    6. Series
       1. Series are a special classification that can be assigned to WCA Competitions held less than 33 days apart for which the distance between them is less than 200 km, with the intention to accommodate more competitors to compete in one area.
          1. Series may be accepted if the need for competitions in an area is so great that the community will benefit from having multiple competitions held in a short period of time.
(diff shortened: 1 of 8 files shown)


documents/policies/external/Competition Requirements.md
```diff
@@ -168,8 +176,7 @@ The WCA Delegate must submit the following information when requesting approval
       1. If there is a deadline, it must not be earlier than the time registration is set to close.
       2. The deadline for adding registered events may be left blank, in which competitors may add registered events up until an event takes place.
       3. Competitors may choose not to compete in an event they are registered for.
-      4. Exception: for competitions with only one event, the deadline for adding registered events may be omitted.
-      5. At the discretion of the WCA Delegate(s), a competitor who is deemed to have made a mistake with the user interface during the registration process (e.g. registering for the inverse list of events that they intended) can be granted an exception to the deadline for adding registered events.
+      4. At the discretion of the WCA Delegate(s), a competitor who is deemed to have made a mistake with the user interface during the registration process (e.g. registering for the inverse list of events that they intended) can be granted an exception to the deadline for adding registered events.
```
CarterKoala, line 179: For the Regs, we generally do not renumber points like this when removing regs or articles. I'm not sure if you have a different history with WCRP, but it is something to consider, since past references can become a pain

Nanush7: I think the document compiler wouldn't let you skip numbers in a markdown list.

dmint789:
It would be good to link the explainer document in the description. Also, someone mentioned that one of the changes still says minimum 8 hours between comp days, so that should be changed to 10 too.

jgouly:
Would be good to clarify if:
```
5. The schedule must include a distinct timeframe for all rounds of official events at the time of announcement.
```
Means you can't add an extra round during a competition (if running early etc)

Bitzerr:
I don't really see how any of these changes are necessary. I don't think formality in competitions is something the WCA should be trying to achieve. The primary reason for going to competitions is to have fun.

Justiniscoding:
While I think most of these changes are positive, the last major change doesn't sit right with me. While it is good to prevent people from competing for too long consecutively, why not make this limit reset at midnight? This new rule change would prevent 24 hour competitions from existing, which may not be common but are very popular and a once in a lifetime opportunity for cubers. 

To keep 24 hour comps allowed, why not change
`iv. Official events at a competition must not run for more than a 16-hour period without a break of 10 hours or more`
to 
`iv. Official events at a competition must not run for more than a 16-hour period without a break of 10 hours or more (limit resets at midnight)`

sushantshah-dev:
I don't see any reason for restricting competition duration. Is it causing any issues to the WCA as an organisation? Is it causing trouble to competitors?

dmint789:
I love the updates! @zekemackay06

KerrieJarman (dismissed):
Approved by a formal Board vote

KerrieJarman:
Additional changes related to line 168 approved by Formal Board vote

nsilvestri (approved):

PR #436, opened 2024-10-17
Fix broken link in WIC motion

dmint789:
The link to the Suspensions and Sanctions motion was outdated. It will now link to [this one](https://documents.worldcubeassociation.org/documents/motions/15.2024.1%20-%20Suspensions%20and%20other%20Sanctions.pdf).


nsilvestri (approved):
approved without Board vote as non-content change

PR #445, opened 2024-12-28
Replace references of "region of representation" with "country"

Sixstringcal:
Replace references of "region of representation" with "country"

Not to be merged until Jan 1.

I'm somewhat surprised this was only used in the Delegate Handbook (unless I missed something).


dmint789 (approved):
LGTM! Won't be merging until the 1st.

PR #444, opened 2024-12-24
Revert "WCRP 5.3 Update (#435)"

nsilvestri:
This reverts commit c5596dcf49d88a762f5b8e2976c4bfc54f9b93c5.

Accidentally merged prematurely, will go live January 1, 2025.

PR #448, opened 2025-01-05
WCRP 5.3 Update

nsilvestri:
Reverts thewca/wca-documents#444, which itself reverts thewca/wca-documents#435, thus making the changes live. 

Already approved by Board vote.


KerrieJarman (approved):
Approved by the Board

PR #450, opened 2025-01-08
Replace Staff with Volunteer

KerrieJarman:
The WCA Board is proposing to transition away from using the term "WCA Staff" and replacing it with "WCA Volunteer" to describe individuals who contribute their time and skills to the WCA in a volunteer capacity. We also propose adopting “Competition Volunteer” where we have used “Competition Staff” in the past.

PR #458, opened 2025-02-08
Add WCA Age Policy v1.0

nsilvestri:
This PR adds a WCA Age Policy.

It must not be merged until after #450 has been merged, due to the "volunteer" language contained within.

### DIFF
--- a/documents/policies/internal/Age.md
+++ b/documents/policies/internal/Age.md
@@ -1,0 +1,19 @@
+# WCA Age Policy
+
+### Version 1.0 {.version}
+
+## Purpose
+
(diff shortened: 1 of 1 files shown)


ohexter:
Having age as a defined term, including years as the unit within 1, but then using "years of age" in 2 is somewhat inelegant, although alternatives such as "has an age of at least X" don't read very naturally either.

nsilvestri:
@ohexter thanks for the note, I have adjusted the language to be "age is calculated as..." rather than "defined as" in light of this.

KerrieJarman (approved):
Approved by unanimous Board vote

Issue #454, opened 2025-01-24
Reword "other potential organizers may not have altruistic intentions" in the Delegate handbook


OWelton-Rosie:
I (and others) have misinterpreted "other potential organizers may not have altruistic intentions". The current wording could be misconstrued as "other potential organizers **might** not have altruistic intentions". 

I'd rephrase this. It's best to leave no doubts that competitions are not-for-profit endeavours.

Located in the 1st paragraph of page 14.

roxanajohnson88:
I think it's fine as is. "May" and "Might" have little difference in meaning beyond the fact that may indicates that something is more likely to happen, which is not something that we should assume.

PR #464, opened 2025-03-03
Minor Clarity Fix in Delegate Handbook

Nanush7:
Closes https://github.com/thewca/wca-documents/issues/454: potential confusion with the use of "may", which is used following RFC 2119 in other documents.


dmint789 (approved):
LGTM!

PR #469, opened 2025-03-14
Update the equipment funding policy to reflect the change in terminology

OWelton-Rosie:
As far as I'm aware, the term is now "WCA Volunteer", not "WCA Staff member", so (at least in my mind), it makes logical sense to change the policy to reflect the new term.


KerrieJarman:
Thanks for notifying us. There is already a PR under review that updates all policies.

OWelton-Rosie:
> Thanks for notifying us. There is already a PR under review that updates all policies.

Sorry about that (🤦 ), I should have checked the PRs. 

Yeah, in that sense it makes perfect sense to close this.

PR #470, opened 2025-03-24
Documents Motion - Clerical Fix

ohexter:
The "Submitted by" section of the Documents Motion contains text that does not make sense in context. All other Motions say "WCA Board".

This text appears to have been included since the Motion was originally created ([PR](https://github.com/thewca/wca-documents/pull/218/files)).

This appears to be a drafting error. I believe it is suitable for correction as a clerical error under the Amendments of Motions Motion.


KerrieJarman (approved):
Approved via Board vote

PR #471, opened 2025-04-03
Align Language and Fix Grammar/Spelling Across Edudoc Files

sarahstrong314:
### Consistency with WCA Website and Documents

- **Terminology**
	- cube cover → puzzle cover
	- puzzle drop-off station → puzzle drop-off table
	- competing area → solving stations
	- competitor waiting area → waiting area
	- scrambling station → scrambling table
	- stationary judges → seated judges
	- non-Delegate organizer → Community organizer
	- registration terms and conditions → registration requirements
	- winner's ceremony → awards ceremony
	- Round 1 → First Round (when referring to a specific first round)
	- round 1 → the first round (when referring to a general first round)
	- +2 → +2 penalty
	- US → USA
	- United States → USA
- **Capitalization**
 	- Delegate Report
 	- Regional Delegate
 	- Regional Organization
 	- Regulations
 	- WCA Team
 	- WCA Committee
 	- Community Member
 	- Extra Registration Requirements

### Consistency with WCA Regulations

- **Terminology**
	- scorecard → score sheet
	- data entry → score-taking (noun or adjective)
	- data entry person → score taker
	- extra → extra attempt
	- provisional extra → provisional extra attempt
	- organizational team → organization team
	- organizer team → organization team
	- organizational staff → organizers
	- member of the organization team → organizer
	- issue/issuing extra attempts → grant/granting extra attempts
	- ex. → e.g.
- **Style**
	- write all event names in full
	- expand contractions
	- avoid sentences that end in "etc."
	- no auto-capitalizing after "e.g."
	- no auto-capitalizing after colons (except "Note:")
	- use curved quotation marks (' → ’, " → “/”, except in HTML code)
	- put punctuation marks outside of quotation marks

### General Consistency

- **Terminology**
	- first-timer → first-time competitor
	- new competitor → first-time competitor
	- new-comer/newcomer → first-time competitor 
 		- This is an important distinction since being a newcomer doesn't imply it's their first time, just that they started competing recently.
	- e-mail → email
	- gmail/GMail → Gmail
	- insurance fee → liability insurance coverage
	- Delegates of the competition → listed Delegates
- **Style**
	- use bold for emphasis instead of upper casing (except for speech)
	- remove colons at the end of headers
	- remove "WCA" from `h1` tags
	- title casing in `h1` and `h2` tags
	- sentence casing in `h3` tags (except in Delegate Handbook)
	- remove `h4` tags that aren't under a `h3` tag
[...]

PR #477, opened 2025-04-28
staff -> volunteer in crash course

pranav-027:
Renamed staff -> volunteers in crash course and disciplinary guidelines


sarahstrong314:
Hi Pranav,

Thanks for doing this! Currently, there is another PR (#471) with propose changes to the Staff/Volunteer Crash Course, so it would be best to make your changes after the first PR is closed. The other edudocs also need the Volunteer terminology updated and it would be efficient to put all the Volunteer terminology changes into one PR. I've closed this PR for now until progress is made with the first PR.

PR #478, opened 2025-05-02
Staff -> Volunteer wording changes in Edudocs

sarahstrong314:
Summary of wording changes:

* Volunteer with upper case "V" for WCA Volunteers
* volunteer with lower case "v" for competition volunteers
* consistent use of "role" to refer to a volunteer position
* consistent use of "duties" to refer to the responsibilities of a volunteer
* consistent use of "assignment" to refer to a specific event+round+group a volunteer is assigned to
* venue staff -> venue employees

Summary of filename changes:

* `edudoc/staff-crash-course/staff-crash-course.md` -> `edudoc/volunteer-crash-course/volunteer-crash-course.md`
* `edudoc/organizer-guidelines/staff.md` -> `edudoc/organizer-guidelines/competition-volunteers.md`

PR #474, opened 2025-04-13
Update Guidelines to Disciplinary Incidents

slongbehn:
Substantive:
- Expanded intro
- Expanded information on Banned Competitors list
- Expanded behavioral guidance
- Expanded information regarding WCA Volunteers
- Reworked for current Delegate Report practices 
- Added general legal checkpoints

Clerical:
- Update "WCA Staff" to "WCA Volunteer"
- Updated links
- Added links
- Grammatical changes
- Reworked sections with 1st person pronouns

PR #484, opened 2025-05-15
Adjust Page Breaks in Edudocs

sarahstrong314:
The changes in #474 and #478 offset the page breaks. I've re-adjusted them accordingly.


dmint789 (approved):
LGTM

PR #485, opened 2025-05-18
Fixed Table of Contents in Delegate Handbook

sarahstrong314:


edudoc/delegate-handbook/delegate-handbook.md
```diff
@@ -1112,7 +1113,7 @@ This section covers the responsibilities of a Delegate outside of directly deleg
 
 ## Managing WCA User Accounts
 
-### Accounts vs. WCA IDs (Profiles) {#accounts-vs-wca-ids}
+### Accounts vs. Profiles (WCA IDs)
```
this-chord, line 1116: Line 92 needs to be changed too then

sarahstrong314: Fixed!

this-chord: Did you check that the link still works though? I don't think it will work with you having deleted the link here. You need {#accounts-vs-profiles} at the end of this line

this-chord:
I'm not quite understanding the indentation changes. It looks like 4 spaces were replaced with 4 spaces, at least in the GitHub diff.

mintydev789 (approved):
LGTM

PR #487, opened 2025-07-08
Committees and Teams Clerical - July 2025

ohexter:
The reference to 4.2 in 5.3.3 should instead refer to 5.2. This appears to be a drafting error, as the correct reference would have been 4.2 in the version prior to the changes being introduced , see https://github.com/thewca/wca-documents/commit/5018d80a85858613340b2161efab0e9c5a7bca6f.


nsilvestri (approved):
Approved by Board vote

PR #508, opened 2025-12-14
Add ED Incentive Compensation policy

nsilvestri:
This PR creates the Executive Director Incentive Compensation policy, outlining the procedures around executive director incentives.


ohexter (commented):
1.1 - this is the only mention of the Compensation Committee; other references elsewhere are only to the Board
1.2 - some potential tension between this mechanism (Board / ED agree) and that in 1.1 (Board approve)
4.1 - unnecessary line break in text (shouldn't impact compiled pdf)

PR #553, opened 2026-04-03
Updated proximity policy to align with WCRP 5.4

LarsJohanFolde:
Changed proximity from 10 km and 5 days to 30 km and 5 days


sarahstrong314 (approved):
LGTM! Thanks, Lars.

PR #560, opened 2026-05-08
Restore previous explanation of unpermitted objects

sarahstrong314:
Restore the explanation of unpermitted objects from #511 that were overwritten in #542


Hfaerden (approved):
LGTM!

2016WUJI01 (approved):
LGTM

PR #564, opened 2026-05-21
Removed outdated sentence from Delegate Handbook

sarahstrong314:
Removed `If a Delegate is not promoted, their promotion is simply postponed until the next promotion cycle (i.e. next quarter).` This is not the current practice as Senior Delegates do not recommend Junior Delegates for promotion without recent input from their Regional Delegate.


MBennisWCA (approved):
LGTM

PR #576, opened 2026-07-02
Add trailing newlines to docs that were missing them

Abhinav-Khot:
Trailing newlines are required for POSIX compliance and also help in producing better `git` Diffs. Most of the docs under `edudoc` adhere to this barring two. This fixes that.


sarahstrong314 (approved):
LGTM! The diff doesn't show the newline but I clicked into the files to view them.

PR #596, opened 2026-09-02
Fixed table and images in Organizer Handbook

sarahstrong314:
The HTML table was missing some tags that GitHub was okay with, but the PDF build was not. The images had the same syntax issue and one filename was incorrect.


abcoolidge (approved):
LGTM

PR #597, opened 2026-09-02
Delete obsolete folders: organizer-guidelines and volunteer-crash-course

sarahstrong314:
Now that the Organizer Handbook and Volunteer Handbook are published and accessible via the WCA Website, we can delete their predecessors so that the old links no longer work.


abcoolidge (approved):
LGTM!

